PR #4, opened 2021-02-23
Add ccd config library

mapingo:
### Change description ###
Add ccd config generator plugin
Add ccd definition processor


package.json
```diff
@@ -0,0 +1,56 @@
...
+  "main": "index.js",
+  "repository": "git@github.com:hmcts/nfdiv-case-api.git",
+  "license": "MIT",
+  "private": true,
+  "engines": {
+    "node": ">=12.0.0"
+  },
+  "config": {
+    "local": {
+      "cosUrl": "http://nfdiv-cos:4012",
```
nitinprabhu, line 14: do we cos now ? callbacks will be handled by this case api library

linusnorton: nope, should be nfdiv-case-api

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "ccd-definition-processor"]
```
linusnorton, line 1: rather than using a submodule we should use the docker image, I'll show you tomorrow

build.gradle
```diff
@@ -78,12 +69,6 @@ task integration(type: Test) {
     failFast = true
 }
 
-task smoke(type: Test) {
```
linusnorton, line 81: I think we need a smoke test

ccd-case-generator/build.gradle
```diff
@@ -0,0 +1,33 @@
+plugins {
```
linusnorton, line 1: We're maintaining the CCD def as a separate module? If so I think it should be ccd-definition or something.

ccd-case-generator/build/tmp/compileJava/source-classes-mapping.txt
```diff
@@ -0,0 +1,8 @@
+uk/gov/hmcts/reform/ccd/ccd/model/UserRole.java
```
linusnorton, line 1: think this needs an exclude?

package.json
```diff
@@ -0,0 +1,19 @@
+{
```
linusnorton, line 1: I wonder if this should be inside the ccd-case-generator module?

package.json
mapingo: I have removed these completely as we can do this in Java rather than when the spreadsheet is created

package.json
mapingo: My thinking too, I think this can be moved inside

ccd-case-generator/build.gradle
mapingo: I went with a sub module as the ccd-config-generator plugin only works with master and this seemed to be the easiest way to exclude any CcdConfig classes.  We could add some configuration to the plaugin to allow a different source directory if we want to simplify things

linusnorton (approved):

PR #7, opened 2021-02-24
Update values.yaml

damongreen123:
**Before creating a pull request make sure that:**

- [ ] commit messages are meaningful and follow good commit message guidelines

Please remove this line and everything above and fill the following sections:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/charts/nfdiv-case-api/values.preview.template.yaml
+++ b/charts/nfdiv-case-api/values.preview.template.yaml
@@ -1,5 +1,5 @@
 tags:
-    ccd-idam-pr: false
+    ccd-idam-pr: true
 
 java:
   # Don't modify below here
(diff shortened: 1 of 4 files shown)


charts/nfdiv-case-api/values.yaml
```diff
@@ -8,3 +8,7 @@ java:
       secrets:
         - AppInsightsInstrumentationKey
   environment:
+ccd:
+  enabled: false
+ccd-idam-pr:
+  enabled: false
```
nitinprabhu, line 14: if idam is not enabled by default should we enable in preview ?

charts/nfdiv-case-api/values.preview.template.yaml
```diff
@@ -1,3 +1,6 @@
+tags:
+    ccd-idam-pr: false
```
nitinprabhu, line 2: has to be true for pr

PR #12, opened 2021-02-26
Add ccd defintions test framework

mapingo:
Add test stub
Add Case Data fields


ccd-definitions/.gitignore
```diff
@@ -29,4 +29,13 @@ yarn-error.log
 
 ### Generated CCD Json Definition ###
 ccd-definition
-definitions/development
```
banderous, line 32: Do you need to version the json at all? could you generate it into the build folder?

ccd-definitions/src/main/java/uk/gov/hmcts/reform/divorce/ccd/event/DraftCreate.java
```diff
@@ -29,14 +29,10 @@ public void buildWith(final ConfigBuilder<CaseData, State, UserRole> builder) {
...
-            .grant("CRU", CASEWORKER_DIVORCE_COURTADMIN_BETA)
-            .grant("CRU", CASEWORKER_DIVORCE_COURTADMIN)
-            .grant("CRU", CASEWORKER_DIVORCE_SOLICITOR)
-            .grant("CRU", CASEWORKER_DIVORCE_SUPERUSER)
-            .grant("CRU", CASEWORKER_DIVORCE_COURTADMIN_LA)
-            .fields()
-            .optional(CaseData::getCreatedDate)
-            .optional(CaseData::getD8caseReference)
-            .optional(CaseData::getD8legalProcess);
+            .grant("R", CASEWORKER_DIVORCE_COURTADMIN_BETA)
```
banderous, line 32: you can pass multiple roles into grant:

grant("R", ROLE_1, ROLE_2...)

Also if you find yourself using the same set of roles in multiple places you could store them in a collection and pass that in to grant.

ccd-definitions/src/main/java/uk/gov/hmcts/reform/divorce/ccd/model/RejectReason.java
```diff
@@ -0,0 +1,34 @@
...
+import uk.gov.hmcts.ccd.sdk.types.ComplexType;
+import uk.gov.hmcts.reform.divorce.ccd.model.enums.RejectReasonList;
+
+import static uk.gov.hmcts.ccd.sdk.types.FieldType.FixedList;
+import static uk.gov.hmcts.ccd.sdk.types.FieldType.TextArea;
+
+@Data
+@Builder
+@AllArgsConstructor
+@ComplexType(generate = true)
```
banderous, line 17: You shouldn't need to use the complex type annotation here, if it's reachable from your case data class then it should be generated.

I was planning on getting rid of the @ComplexType annotation and just having @CCD

ccd-definitions/src/main/java/uk/gov/hmcts/reform/divorce/ccd/model/RejectReason.java
```diff
@@ -0,0 +1,34 @@
...
+public class RejectReason {
+
+    @JsonProperty("RejectReasonType")
+    @CCD(
+        label = "Reject reason",
+        type = FixedList,
```
banderous, line 23: You should only need to provide label here, type and typeParameter should be inferred automatically.

PR #15, opened 2021-03-01
update ccd definitions lib

linusnorton:


nitinprabhu (approved):

nitinprabhu (commented):
There is already one created by dependabot https://github.com/hmcts/nfdiv-case-api/pull/14

PR #16, opened 2021-03-01
Added script to generate and upload ccd definitions to preview and AAT

nitinprabhu:
- https://tools.hmcts.net/jira/browse/NFDIV-502
- Added scripts to generate ccd definition xls using `definition-processor` docker image.
- Also script uploads xls to preview and AAT environment.
- Currently json files are not generated and script uses predefined json which are picked from ccd definition repository.
- While reviewing please ignore the json files as they would be removed later and we would using json files generated from ccd library.


Jenkinsfile_CNP
```diff
@@ -10,10 +10,72 @@ def component = "case-api"
...
+                secret('microservicekey-ccd-gw', 'API_GATEWAY_S2S_KEY'),
+                secret('microservicekey-ccd-admin', 'ADMIN_S2S_KEY'),
+                secret('microservicekey-ccd-gw', 'CCD_API_GATEWAY_S2S_SECRET'),
+                secret('microservicekey-ccd-data', 'CCD_DATA_STORE_S2S_SECRET'),
+                secret('microservicekey-ccd-definition', 'CCD_DEFINITION_STORE_S2S_SECRET')
+        ],
+        'ccd-${env}'      : [
+                secret('ccd-api-gateway-oauth2-client-secret', 'API_GATEWAY_IDAM_SECRET'),
+                secret('ccd-admin-web-oauth2-client-secret', 'ADMIN_WEB_IDAM_SECRET'),
+                secret('postcode-info-address-lookup-token', 'ADDRESS_LOOKUP_TOKEN'),
```
linusnorton, line 34: do we need this?

bin/s2s-token.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -eu
+
+microservice=${1}
+oneTimePassword=${2}
+
+curl --insecure --fail --show-error --silent -X POST \
+  ${SERVICE_AUTH_PROVIDER_API_BASE_URL:-http://localhost:4502}/lease \
```
linusnorton, line 9: Rather than using the docker container to generate the OTP, you can just call 
```
token=$(curl $CURL_OPTS -X POST \
  -H "Content-Type: application/json" \
  -d '{"microservice":"'${MICROSERVICE}'"}' \
  http://rpe-service-auth-provider-aat.service.core-compute-aat.internal/testing-support/lease)
``` 
to get a token

linusnorton (approved):

mapingo (approved):

Jenkinsfile_CNP
nitinprabhu: yes used here https://github.com/hmcts/nfdiv-case-api/blob/master/charts/nfdiv-case-api/values.preview.template.yaml#L91

bin/s2s-token.sh
nitinprabhu: sure will remove docker and use testing support

nitinprabhu: done

Jenkinsfile_CNP
linusnorton: eeesh, we shouldn't be loading in vault values that way... but if that is how everyone does it then we can do it the same

PR #49, opened 2021-03-08
Fix sonar reporting

mapingo:
### Change description ###
Fix sonar coverage for sub modules
Fix Checkstyle and PMD
Fix PMD code issues
Fix Sonar code issues
Update Jenkins pipeline

```
[ ] Yes
[ x] No
```


build.gradle
```diff
@@ -23,3 +24,17 @@ dependencyCheck {
...
+]
+
+sonarqube {
+    properties {
+        property "sonar.projectName", "Reform :: nfdiv-case-api"
+        property "sonar.projectKey", "uk.gov.hmcts.reform:nfdiv-case-api"
+        property "sonar.coverage.jacoco.xmlReportPaths", "${projectDir}/ccd-definitions/build/reports/jacoco/test/jacocoTestReport.xml,${projectDir}/api/build/reports/jacoco/test/jacocoTestReport.xml"
+        property "sonar.exclusions", sonarExclusions.join(', ')
+        property "sonar.login", "admin"
+        property "sonar.password", "8Y573ccS2mpwWFX"
```
nitinprabhu, line 38: username/password ? need to remove this

nitinprabhu: ok discussed on slack it is local credentials :)

Jenkinsfile_CNP
```diff
@@ -49,6 +49,9 @@ withPipeline(type, product, component) {
 
     after('test') {
         builder.gradle('integration')
+        builder.gradle('jacocoTestReport')
```
nitinprabhu, line 52: do we need to run Jacoco explicitly ? sonarqube task has a dependency on jacoco

mapingo: It looks like this needs to be run explicitly.  Jacoco refuses to run otherwise because there is no src folder in the root

nitinprabhu (approved):

Jenkinsfile_CNP
```diff
@@ -49,6 +49,9 @@ withPipeline(type, product, component) {
...
+        steps.junit '**/test-results/**/*.xml'
+        steps.archiveArtifacts allowEmptyArchive: true, artifacts: '**/build/reports/**/*'
```
nitinprabhu, line 54: do we need this archiving here ?

mapingo: removed

PR #60, opened 2021-03-10
use reflections to detect CCD config

linusnorton:
Use reflections to detect CCD config


ccd-definitions/src/test/java/uk/gov/hmcts/reform/divorce/ccd/event/PatchCaseTest.java
```diff
@@ -57,7 +57,7 @@ void shouldBuildPatchCaseEventWithConfigBuilder() {
             CASEWORKER_DIVORCE_SUPERUSER,
             CASEWORKER_DIVORCE_COURTADMIN_LA);
         verify(eventBuilder, times(1)).fields();
-        verify(fieldCollectionBuilder, times(21)).optional(any());
+        verify(fieldCollectionBuilder, times(24)).optional(any());
```
nitinprabhu, line 60: this is available in latest master :)

nitinprabhu (approved):

ccd-definitions/src/main/java/uk/gov/hmcts/reform/divorce/ccd/DevelopmentCcdConfig.java
```diff
@@ -9,13 +11,18 @@
...
 
+    @SneakyThrows
     @Override
     public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuilder) {
-
         configBuilder.setEnvironment("development");
         configBuilder.setWebhookConvention(this::webhookConvention);
 
-        new CcdConfigApplier(new NoFaultDivorceCcdConfigFactory()).applyTo(configBuilder);
+        Reflections reflections = new Reflections("uk.gov.hmcts.reform.divorce.ccd");
```
mapingo, line 20: The reason I setup a Factory for the configuration was to make it easy to add other Case types in the future.  We may need to revert to that method or be more creative with the package names if we use this reflection method.

linusnorton: that's fair enough, I think it might be best to have a whole other gradle project for that

PR #64, opened 2021-03-11
Integrate s2s for only allowing configured service to invoke callback

nitinprabhu:
### Change description ###

- Integrated S2S to verify if service is allowed to invoked save and signout callback.
- Includes unit, integration and functional tests.
- Tested locally.
- Includes minor refactoring to unit test for controller.

```
[ ] Yes
[ x] No
```


api/src/functionalTest/java/uk/gov/hmcts/reform/divorce/caseapi/SaveAndSignOutNotificationTest.java
```diff
@@ -32,13 +37,20 @@
     @Value("${test-url}")
     private String testUrl;
 
+    @Value("${idam.s2s-auth.url}")
+    private String s2sUrl;
```
linusnorton, line 41: At some point we might need to move all this out to some kind of helper to avoid boilerplate, but it's fine for now

api/src/functionalTest/java/uk/gov/hmcts/reform/divorce/caseapi/SaveAndSignOutNotificationTest.java
```diff
@@ -120,6 +134,56 @@ public void shouldFailValidationErrorWhenEmailIsMissing() {
...
+                    .build()
+            )
+            .when()
+            .post(SAVE_AND_SIGN_OUT_CALLBACK_URL);
+
+        assertThat(response.getStatusCode()).isEqualTo(UNAUTHORIZED.value());
+    }
+
+    @Test
+    public void shouldReturn403WhenServiceIsNotAuthorised() {
```
linusnorton, line 163: as with the previous comment we can probably have a generic check for 403 response rather than doing it on every single endpoint

api/src/main/java/uk/gov/hmcts/reform/divorce/caseapi/controllers/SaveAndCloseController.java
```diff
@@ -24,14 +30,32 @@
...
+        @ApiResponse(code = 403, message = "Service not configured to invoke callback")
+    })
+    public void saveAndClose(
+        @RequestHeader("ServiceAuthorization") String serviceAuthToken,
+        @RequestBody CcdCallbackRequest callbackRequest
+    ) {
         log.info("Save and sign out callback invoked");
 
+        String serviceName = tokenValidator.getServiceName(serviceAuthToken);
+        if (!authorisedServices.contains(serviceName)) {
```
linusnorton, line 54: Can we move all this into a higher level check for every request?

charts/nfdiv-case-api/values.yaml
```diff
@@ -18,3 +18,5 @@ java:
         - uk-gov-notify-api-key
   environment:
       NOTIFY_TEMPLATE_SIGN_IN_URL: https://nfdiv.{{ .Values.global.environment }}.platform.hmcts.net/
+      S2S_URL: http://rpe-service-auth-provider-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal
+      S2S_AUTHORISED_SERVICES: ccd_data
```
linusnorton, line 22: Let's not set this if the default works fine

linusnorton (approved):

PR #103, opened 2021-03-26
change from set to single as frontend now only returns 1 value

velimd:
https://tools.hmcts.net/jira/browse/NFDIV-425

### Change description ###
The old frontend had this property has an Set as it allowed users to have more than one option but the new implementation allows users to only select one option.

```
[ ] Yes
[X] No
```

### DIFF
--- a/ccd-definitions/src/main/java/uk/gov/hmcts/reform/divorce/ccd/model/CaseData.java
+++ b/ccd-definitions/src/main/java/uk/gov/hmcts/reform/divorce/ccd/model/CaseData.java
@@ -220,7 +220,7 @@ public class CaseData {
     @CCD(
         label = "How did the petitioner change their name?"
     )
-    private Set<ChangedNameHow> d8PetitionerNameChangedHow;
+    private ChangedNameHow d8PetitionerNameChangedHow;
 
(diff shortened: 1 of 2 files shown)


ccd-definitions/src/main/java/uk/gov/hmcts/reform/divorce/ccd/model/CaseData.java
```diff
@@ -220,7 +220,7 @@
     @CCD(
         label = "How did the petitioner change their name?"
     )
-    private Set<ChangedNameHow> d8PetitionerNameChangedHow;
+    private ChangedNameHow d8PetitionerNameChangedHow;
```
nitinprabhu, line 223: If you change field type then elastic search will be broken.
Think you need to bump the case type  too with this PR https://github.com/hmcts/nfdiv-case-api/blob/master/ccd-definitions/src/main/java/uk/gov/hmcts/reform/divorce/ccd/ccdcase/NoFaultDivorce.java#L22 to `NO_FAULT_DIVORCE4`.
Other option is to ask platops to delete ES indexes and then this should be fine.

nitinprabhu (approved):

PR #150, opened 2021-04-14
Rename petitioner to applicant

mapingo:
### Change description ###
Change petitioner to applicant across code base


linusnorton (commented):
It looks fine but we'll need another for respondent to co-applicant. Don't pull the trigger yet, I believe we're awaiting approval from policy

PR #183, opened 2021-04-23
feat(NFDIV-305): Added documents field

SebFlippence:

[NFDIV-305](https://tools.hmcts.net/jira/browse/NFDIV-305) Upload documents

### Change description ###

Added fields for supporting documents and cannot upload documents.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
+++ b/src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
@@ -577,6 +577,21 @@ public class CaseData {
     private Set<ClaimsCostFrom> divorceClaimFrom;
 
     @CCD(
+        label = "Supporting documents",
+        hint = "Such as marriage/civil union certificate, proof of name changes, certified translations etc.",
+        typeOverride = TextArea,
+        access = {DefaultAccess.class}
+    )
+    private String supportingDocuments;
+
+
+    @CCD(
+        label = "Cannot upload supporting documents",
+        access = {DefaultAccess.class}
+    )
+    private Set<SupportingDocuments> cannotUploadDocuments;
+
+    @CCD(
         label = "Created date",
         hint = "Date case was created",
         access = {DefaultAccess.class}
--- a/src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
+++ b/src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
@@ -577,19 +577,19 @@ public class CaseData {
     private Set<ClaimsCostFrom> divorceClaimFrom;
 
     @CCD(
-        label = "Supporting documents",
+        label = "Uploaded supporting documents",
         hint = "Such as marriage/civil union certificate, proof of name changes, certified translations etc.",
         typeOverride = TextArea,
         access = {DefaultAccess.class}
     )
-    private String supportingDocuments;
+    private String uploadedDocuments;
 
 
     @CCD(
         label = "Cannot upload supporting documents",
         access = {DefaultAccess.class}
     )
-    private Set<SupportingDocuments> cannotUploadDocuments;
+    private Set<SupportingDocumentType> cannotUploadDocuments;
 
     @CCD(
         label = "Created date",
--- a/src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
+++ b/src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
@@ -579,17 +579,18 @@ public class CaseData {
     @CCD(
         label = "Uploaded supporting documents",
         hint = "Such as marriage/civil union certificate, proof of name changes, certified translations etc.",
-        typeOverride = TextArea,
+        typeOverride = Collection,
+        typeParameterOverride = "DivorceDocument",
         access = {DefaultAccess.class}
     )
-    private String uploadedDocuments;
+    private List<DivorceDocument> supportingDocumentMetadata;
 
 
     @CCD(
         label = "Cannot upload supporting documents",
         access = {DefaultAccess.class}
     )
-    private Set<SupportingDocumentType> cannotUploadDocuments;
+    private Set<SupportingDocumentType> cannotUploadSupportingDocument;
 
     @CCD(
         label = "Created date",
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
```diff
@@ -576,6 +576,21 @@
     )
     private Set<ClaimsCostFrom> divorceClaimFrom;
 
+    @CCD(
+        label = "Supporting documents",
+        hint = "Such as marriage/civil union certificate, proof of name changes, certified translations etc.",
+        typeOverride = TextArea,
+        access = {DefaultAccess.class}
+    )
+    private String supportingDocuments;
```
tomxelliott, line 585: How comes this field is of type `String` rather than being a Set of `SupportingDocuments` like the `cannotUploadDocuments` field?

tomxelliott: Is it because it is an enum?

src/main/java/uk/gov/hmcts/divorce/common/model/SupportingDocuments.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.hmcts.divorce.common.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import uk.gov.hmcts.ccd.sdk.api.HasLabel;
+
+@Getter
+@AllArgsConstructor
+public enum SupportingDocuments implements HasLabel {
```
linusnorton, line 10: could we use singlar class names, ie `SupportingDocument`. And perhaps this class should be `SupportingDocumentType` if it's not the document itself

src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
SebFlippence: This will actually be metadata about the uploaded documents, e.g. id, url, name, upload time etc.

tomxelliott: ok that makes sense. Will this String field hold information/metadata about multiple documents? If so, how can we differentiate between the different documents?

nitinprabhu: are you sure document metadata is represented as text(String) in ccd ? There is complex type for this not sure how this will look in CCD

SebFlippence: @tomxelliott there can be multiple documents uploaded, for now I was going to store this as a JSON array, but if there's a complex Java type for documents uploaded to https://github.com/hmcts/document-management-client then it might be better to use that. @nitinprabhu Do you have a link to the type or an example of it being used?

tomxelliott: @SebFlippence there is a Document type in CCD

nitinprabhu: https://github.com/hmcts/nfdiv-case-api/blob/master/src/main/java/uk/gov/hmcts/divorce/document/model/DivorceDocument.java

I think you can use this but make sure you have all the fields

src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
```diff
@@ -576,6 +576,21 @@
     )
     private Set<ClaimsCostFrom> divorceClaimFrom;
 
+    @CCD(
+        label = "Uploaded supporting documents",
+        hint = "Such as marriage/civil union certificate, proof of name changes, certified translations etc.",
+        typeOverride = TextArea,
+        access = {DefaultAccess.class}
+    )
+    private String uploadedDocuments;
```
velimd, line 585: this is interesting, this should either be the set of documents uploaded or a yes or no for confirmation of document uploaded. Thats what im getting form the field name.

src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
```diff
@@ -576,6 +576,21 @@
...
+
+
+    @CCD(
+        label = "Cannot upload supporting documents",
+        access = {DefaultAccess.class}
+    )
+    private Set<SupportingDocumentType> cannotUploadDocuments;
```
velimd, line 592: this seems like it should be a boolean or text to inform why they cant.

velimd (approved):

src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
SebFlippence: 👍 I've updated this to be:
```java
private List<DivorceDocument> supportingDocumentMetadata;
```

src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
SebFlippence: This will be list of documents that they can't upload:

Screenshot 2021-04-23 at 15 18 17

src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
nitinprabhu: would be nice to test it locally(if you can) and see if it works as expected

src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
```diff
@@ -576,6 +576,22 @@
     )
     private Set<ClaimsCostFrom> divorceClaimFrom;
 
+    @CCD(
+        label = "Uploaded supporting documents",
+        hint = "Such as marriage/civil union certificate, proof of name changes, certified translations etc.",
+        typeOverride = Collection,
```
linusnorton, line 582: why are the overrides needed?

linusnorton (approved):

src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
nitinprabhu: I think for collection it is required for ccd to know the type

PR #184, opened 2021-04-23
Add optional middle name fields

lara-mpt:
### Change description ###

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/divorce/citizen/notification/SaveAndSignOutNotificationHandler.java
```diff
@@ -36,6 +37,7 @@
     public void notifyApplicant(CaseData caseData) {
         Map<String, String> templateVars = new HashMap<>();
         templateVars.put(FIRST_NAME, caseData.getPetitionerFirstName());
+        templateVars.put(MIDDLE_NAME, caseData.getPetitionerMiddleName());
```
nitinprabhu, line 40: do you display middle name in the notify email ? And is the gov notify email template updated ?

lara-mpt: Oh right, I don't think so. At least not yet

nitinprabhu (approved):

PR #188, opened 2021-04-27
Submit notification

linusnorton:
This PR adds the submit notification. There is more work to be done in the submit callback to progress the case state in a subsequent PR.


src/main/java/uk/gov/hmcts/divorce/citizen/event/PaymentMade.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Autowired
+    private ApplicationSubmittedNotification notification;
+
+    @Override
+    public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuilder) {
+        configBuilder
+            .event(PAYMENT_MADE)
+            .initialState(Draft)
+            .postState(Set.of(Submitted))
```
nitinprabhu, line 36: do you want to change the state by default to submitted or do you want to leave this to * and then in the callback change it to submitted or awaiting payment depending on the outcome ?

linusnorton: Yes that would be more sensible

src/main/java/uk/gov/hmcts/divorce/citizen/notification/ApplicationSubmittedNotification.java
```diff
@@ -0,0 +1,47 @@
...
+    private static final DateTimeFormatter formatter = ofPattern("d MMMM yyyy");
+
+    @Autowired
+    private NotificationService notificationService;
+
+    @Autowired
+    private CommonContent commonContent;
+
+    public void send(CaseData caseData, Long id) {
+        Map<String, String> templateVars = new HashMap<>();
```
nitinprabhu, line 30: some logging would be nice here like which notification is sent for which case id

nitinprabhu (approved):

PR #194, opened 2021-04-28
Create dummy payment and add to callback

tomxelliott:


src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorStatementOfTruthPaySubmit.java
```diff
@@ -81,6 +87,15 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
             details.getId()
         );
 
+        log.info("Setting dummy payment to mock payment process");
+        if (caseData.getPayments() == null || caseData.getPayments().isEmpty()) {
+            List<ListValue<Payment>> payments = new ArrayList<>();
+            payments.add(solicitorSubmitPetitionService.getDummyPayment(orderSummary));
+            caseData.setPayments(payments);
+        } else {
+            caseData.getPayments().add(solicitorSubmitPetitionService.getDummyPayment(orderSummary));
```
linusnorton, line 96: so if there is a payment in the list we will add another one?

src/main/java/uk/gov/hmcts/divorce/solicitor/service/SolicitorSubmitPetitionService.java
```diff
@@ -57,4 +60,22 @@ public OrderSummary getOrderSummary() {
             )
             .build();
     }
+
+    public ListValue<Payment> getDummyPayment(OrderSummary orderSummary) {
```
linusnorton, line 64: it kind of feels that this thing should return a Payment and then the callback handler should add the ListValue wrapper. I know it's only temporary but in theory the Payment might not always be used in a list

src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorStatementOfTruthPaySubmit.java
tomxelliott: Yeah, from what I can gather from other use services use cases, if a payment were to fail for one reason or another we would want to store it in our payments array for future records - so any subsequent payments would be added to the payments array and we have a record of attempted and successful payments on the case.

src/main/java/uk/gov/hmcts/divorce/solicitor/service/SolicitorSubmitPetitionService.java
tomxelliott: 👍 - was in two minds about the return type for this one - I agree that returning a Payment makes the most sense

src/main/java/uk/gov/hmcts/divorce/payment/model/Payment.java
```diff
@@ -0,0 +1,75 @@
...
+
+@Data
+@Builder
+public class Payment {
+
+    @CCD(
+        label = "Payment Date",
+        access = {DefaultAccess.class}
+    )
+    @JsonProperty("PaymentDate")
```
linusnorton, line 22: weren't we going to remove all these?

src/main/java/uk/gov/hmcts/divorce/payment/model/Payment.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.time.LocalDate;
+
+@Data
+@Builder
+public class Payment {
+
+    @CCD(
+        label = "Payment Date",
+        access = {DefaultAccess.class}
```
linusnorton, line 20: Do fields on a complex type have access control or is it only the complex type as a whole?

src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
```diff
@@ -724,6 +725,14 @@
     )
     private Set<FinancialOrderFor> financialOrderFor;
 
+    @CCD(
+        label = "Payments",
+        typeOverride = Collection,
```
linusnorton, line 730: were the overrides still needed?

src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorStatementOfTruthPaySubmit.java
```diff
@@ -80,6 +87,23 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
...
+            List<ListValue<Payment>> payments = new ArrayList<>();
+            payments.add(
+                ListValue
+                    .<Payment>builder()
+                    .value(solicitorSubmitPetitionService.getDummyPayment(orderSummary))
+                    .build());
+            caseData.setPayments(payments);
+        } else {
+            caseData.getPayments().add(
+                ListValue
```
linusnorton, line 101: isn't there a `new ListValue(item)`? We might want a static helper method in future: `listValue(item)`

src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorStatementOfTruthPaySubmit.java
```diff
@@ -110,6 +134,20 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
             .build();
     }
 
+    public SubmittedCallbackResponse submitted(CaseDetails<CaseData, State> details,
+                                               CaseDetails<CaseData, State> beforeDetails) {
+        final CaseData caseData = details.getData();
+        final PaymentStatus paymentStatus = caseData.getPayments().get(caseData.getPayments().size() - 1).getValue().getPaymentStatus();
```
linusnorton, line 140: Feels like we might need something a bit more robust here? Sum the total of successful payments and make sure it is the same value as the Fee?

src/main/java/uk/gov/hmcts/divorce/payment/model/Payment.java
tomxelliott: Yep, completely forgot to remove these. well spotted!

linusnorton (approved):

PR #206, opened 2021-04-30
Add send solicitor email notification for solicitor submit

mapingo:
https://tools.hmcts.net/jira/browse/NFDIV-641

### Change description ###
Add send applicant solicitor email notification for solicitor submit


src/main/java/uk/gov/hmcts/divorce/solicitor/service/notification/SolicitorSubmittedNotification.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    private ImmutablePair<EmailTemplateName, String> getSolicitorEmailTemplate(final CaseData caseData) {
+
+        if (caseData.hasPreviousCaseId()) {
+            return ImmutablePair.of(SOL_APPLICANT_SOLICITOR_AMENDED_APPLICATION_SUBMITTED, SENDING_AMENDED_APPLICATION_EMAIL);
+        }
+
+        return ImmutablePair.of(SOL_APPLICANT_SOLICITOR_APPLICATION_SUBMITTED, SENDING_APPLICATION_EMAIL);
+    }
```
nitinprabhu, line 66: I think this code is common as saw it for previous PR. Do we want this to move to a common class rather than repeating ?

mapingo: I have messed around with these two classes to see if there is anything common we can pull into another class, but I just end up making it more complicated.  I have pushed a simplified version to make it clear what is occurring in this component.

nitinprabhu (approved):

PR #209, opened 2021-05-04
NFDIV-608: Rename fields

tomxelliott:
https://tools.hmcts.net/jira/browse/NFDIV-608


src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
```diff
@@ -564,10 +566,10 @@
         label = "The Respondent's home address",
         hint = "If the respondent is to be served at their home address, enter the home address here and as the service "
             + "address below",
-        typeOverride = TextArea,
+        typeOverride = Address,
         access = {DefaultAccess.class}
     )
-    private String respondentHomeAddress;
+    private AddressGlobalUK respondentHomeAddress;
```
nitinprabhu, line 572: as you are changing type you will need to bump case type from nofaultdivorce 6 to 7 otherwise elasticsearch will break

src/main/java/uk/gov/hmcts/divorce/document/content/DraftPetitionTemplateContent.java
```diff
@@ -101,9 +101,19 @@
...
+            respondentPostalAddress = caseData.getDerivedRespondentSolicitorAddr();
+        } else {
+            respondentPostalAddress =
+                caseData.getRespondentHomeAddress().getAddressLine1() + "\n"
+                + caseData.getRespondentHomeAddress().getAddressLine2() + "\n"
+                + caseData.getRespondentHomeAddress().getAddressLine3() + "\n"
+                + caseData.getRespondentHomeAddress().getPostTown() + "\n"
+                + caseData.getRespondentHomeAddress().getCounty() + "\n"
+                + caseData.getRespondentHomeAddress().getPostCode() + "\n"
+                + caseData.getRespondentHomeAddress().getCountry() + "\n";
```
nitinprabhu, line 115: can use StringJoiner here

nitinprabhu (commented):
Also worth testing this on local docker to make sure everything works fine

src/main/java/uk/gov/hmcts/divorce/document/content/DraftPetitionTemplateContent.java
tomxelliott: Used streams instead of StringJoiner in the end as StringJoiner didn't have a way of dealing with null values in an elegant way.

nitinprabhu (approved):

PR #212, opened 2021-05-04
Not sure if we need the serialiser / deserialiser

linusnorton:


nitinprabhu (commented):
I think without the Serializer/Deserializer date was not getting formatted correctly.
If the functional test works after removing we will know :)

PR #214, opened 2021-05-05
Nfdiv 657 remove created date field

mapingo:
https://tools.hmcts.net/jira/browse/NFDIV-657

### Change description ###
Remove createdDate field from CaseData
Use CCD metadata created_date where date is required


linusnorton (approved):

src/main/java/uk/gov/hmcts/divorce/document/DocAssemblyService.java
```diff
@@ -37,11 +38,12 @@
     public DocumentInfo renderDocument(
         CaseData caseData,
         Long caseId,
+        LocalDateTime createDate,
```
nitinprabhu, line 41: this is date time and not date. Do we know how this is going to be shown on the draft petition document ?
I think it should be just date

mapingo: I pushed a commit that makes it clear that LocalDate is used

nitinprabhu (approved):

PR #226, opened 2021-05-06
Remove draft petition after case submission

nitinprabhu:

https://tools.hmcts.net/jira/browse/NFDIV-644

### Change description ###

- Includes functionality to remove draft petition after case submission.
- Includes unit and integration tests.
- Tested end to end on local docker.
- Includes fix when payment is not present else about to submit will fail to complete.

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorStatementOfTruthPaySubmit.java
```diff
@@ -138,7 +139,8 @@ private void updateRespondentDigitalDetails(CaseData caseData) {
     public SubmittedCallbackResponse submitted(CaseDetails<CaseData, State> details,
                                                CaseDetails<CaseData, State> beforeDetails) {
         final CaseData caseData = details.getData();
-        final int feesPaid = caseData.getPayments().stream()
+        final List<ListValue<Payment>> payments = caseData.getPayments() == null ? emptyList() : caseData.getPayments();
```
mapingo, line 142: Is this change covered by a unit test?

src/test/java/uk/gov/hmcts/divorce/solicitor/service/DraftPetitionRemovalServiceTest.java
```diff
@@ -0,0 +1,229 @@
...
+            userId,
+            documentUuid,
+            true
+        );
+
+        verifyNoMoreInteractions(idamService, authTokenGenerator, documentManagementClient);
+    }
+
+    @Test
+    public void shouldThrow403ForbiddenWhenServiceIsNotWhitelistedInDocManagement() {
```
mapingo, line 112: Would it be better to test the 403 and 401 exceptions at the integration level?

src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorStatementOfTruthPaySubmit.java
nitinprabhu: it was existing code which was failing while testing on docker. I think there is a test when payments is present let me add a test when there are no payments

mapingo (approved):

src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorStatementOfTruthPaySubmit.java
nitinprabhu: this is done now

PR #248, opened 2021-05-18
Move validation of jurisdiction into Jurisdiction model

linusnorton:


src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
```diff
@@ -797,4 +746,12 @@ public void addToDocumentsGenerated(final ListValue<DivorceDocument> listValue)
...
         }
     }
+
+    public Jurisdiction getJurisdiction() {
+        if (jurisdiction == null) {
+            jurisdiction = new Jurisdiction();
+        }
+
+        return jurisdiction;
+    }
```
nitinprabhu, line 756: do we need this ?

nitinprabhu (commented):
should we bump the case type as the field names have been changed ?

linusnorton:
> should we bump the case type as the field names have been changed ?

I will double check against the frontend but they should be the same

src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
linusnorton: I tried to set a default in the property definition but it causes issues with lombok so this was the only way I could see if ensuring there was an object

nitinprabhu (approved):

PR #265, opened 2021-05-20
Replaced case data map with model object

nitinprabhu:


linusnorton (approved):
Brilliant - thank you

PR #266, opened 2021-05-21
Move service logic to handler as it is more appropriate in handler

nitinprabhu:


linusnorton (approved):
Personally I think that's better, there's not much code in those services so it may as well go in the handler

PR #279, opened 2021-05-25
Rename events

linusnorton:
New event names:

```
citizen-create-application
citizen-update-application
citizen-save-and-close
citizen-add-payment
citizen-submit-application

solicitor-create-application
solicitor-update-application
solicitor-submit-application
```


nitinprabhu (approved):
Is is worth bumping case type ? all cases created before this may not work as the event names have changed and would result in 404

linusnorton:
> Is it worth bumping case type ? all cases created before this may not work as the event names have changed and would result in 404



> Is it worth bumping case type ? all cases created before this may not work as the event names have changed and would result in 404

Personally I don't think so. The event names aren't built into the cases they're built into the case definition and the application code, so once the definition and the app are up to date things will continue to work

PR #281, opened 2021-05-25
NFDIV-688

tomxelliott:
https://tools.hmcts.net/jira/browse/NFDIV-688

Screenshot 2021-05-25 at 16 40 25


nitinprabhu (commented):
no changes to integration tests ? May be we are not setting court case details in int tests. Worth adding those to int tests requests and verifying the response

nitinprabhu (approved):

PR #291, opened 2021-05-26
NFDIV-688: Fix serialization issue with caseRelatesTo field

tomxelliott:


src/main/java/uk/gov/hmcts/divorce/common/model/LegalProceeding.java
```diff
@@ -18,20 +17,23 @@
...
 
     @CCD(
         label = "Case relates to",
         typeOverride = FixedList,
-        typeParameterOverride = "LegalProceedingsRelated"
+        typeParameterOverride = "LegalProceedingsRelated",
+        access = {DefaultAccess.class}
     )
-    private Set<LegalProceedingsRelated> caseRelatesTo;
+    private LegalProceedingsRelated caseRelatesTo;
```
nitinprabhu, line 31: you would need to bump case type as the field type has changed

nitinprabhu (approved):

PR #292, opened 2021-05-26
Update case type

lara-mpt:
### Change description ###
Update the case type, due to changed fields

```
[ ] Yes
[x] No
```


nitinprabhu (approved):

nitinprabhu (commented):
may not be needed as the other PR is bumping it https://github.com/hmcts/nfdiv-case-api/pull/291

PR #296, opened 2021-05-27
Added validate hwf code event with about to submit callback

nitinprabhu:

https://tools.hmcts.net/jira/browse/NFDIV-588

### Change description ###

- Added validate HWF code event with page controls.
- Added about to submit to set state as per the condition.
- Includes fix for integration tests which was failing due to Jackson not supporting Java 8 date/time type by default.
- Added integration tests
- Tested on local docker

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/integrationTest/java/uk/gov/hmcts/divorce/citizen/event/CitizenAddPaymentTest.java
+++ b/src/integrationTest/java/uk/gov/hmcts/divorce/citizen/event/CitizenAddPaymentTest.java
@@ -1,6 +1,8 @@
 package uk.gov.hmcts.divorce.citizen.event;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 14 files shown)


src/integrationTest/java/uk/gov/hmcts/divorce/citizen/event/CitizenAddPaymentTest.java
```diff
@@ -55,6 +57,8 @@
 @AutoConfigureMockMvc
 public class CitizenAddPaymentTest {
 
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
mapingo, line 60: you could use new ObjectMapper().findAndRegisterModules(), then there would be no need for the setUp method below

nitinprabhu: as discussed still getting error after using `new ObjectMapper().findAndRegisterModules(),` hence will keep this as it is

mapingo (approved):

PR #300, opened 2021-05-28
NFDIV-695 Rename joint jurisdiction connection

lara-mpt:
https://tools.hmcts.net/jira/browse/NFDIV-695

### Change description ###
Renamed the joint jurisdiction connection for clarity

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/ccd/NoFaultDivorce.java
+++ b/src/main/java/uk/gov/hmcts/divorce/ccd/NoFaultDivorce.java
@@ -20,7 +20,7 @@ import static uk.gov.hmcts.divorce.common.model.access.Permissions.READ;
 @Component
 public class NoFaultDivorce implements CCDConfig<CaseData, State, UserRole> {
 
-    public static final String CASE_TYPE = "NO_FAULT_DIVORCE15";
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/divorce/ccd/NoFaultDivorce.java
```diff
@@ -20,7 +20,7 @@
 @Component
 public class NoFaultDivorce implements CCDConfig<CaseData, State, UserRole> {
 
-    public static final String CASE_TYPE = "NO_FAULT_DIVORCE14";
+    public static final String CASE_TYPE = "NO_FAULT_DIVORCE15";
```
nitinprabhu, line 23: don't think we need to bump this

lara-mpt: Ah right, I must be confused. When do we need to bump it? I thought it was when we created or updated the fields

nitinprabhu: if you change the field type then you will need to update

nitinprabhu (approved):

PR #310, opened 2021-06-01
check if user applied for hwf and move to that state if they did.

velimd:
https://tools.hmcts.net/jira/browse/NFDIV-739

### Change description ###

```
[ ] Yes
[X] No
```


src/functionalTest/java/uk/gov/hmcts/divorce/citizen/CitizenSubmitApplicationTest.java
```diff
@@ -42,4 +43,20 @@ public void shouldPassValidationAndGiveSuccessWhenCaseDataValid() throws IOExcep
             .isEqualTo(json(expectedResponse(RESPONSE)));
     }
 
+    @Test
+    public void shouldPassValidationAndGiveSuccessWhenCaseDataValid2() throws IOException {
```
mapingo, line 47: The test name could be changed to shouldPassValidationAndGiveSuccessWhenHelpWithFeesApplied

src/functionalTest/java/uk/gov/hmcts/divorce/citizen/CitizenSubmitApplicationTest.java
```diff
@@ -42,4 +43,20 @@ public void shouldPassValidationAndGiveSuccessWhenCaseDataValid() throws IOExcep
...
+        Map<String, Object> request = caseData(REQUEST);
+        request.put("marriageDate", LocalDate.now().minus(1, YEARS).minus(1, DAYS));
```
nitinprabhu, line 49: is there a reason we need to minus 1 year and 1 day for marriage date ?

src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenSubmitApplication.java
```diff
@@ -81,9 +75,25 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
                 .state(Draft)
                 .build();
         }
+
+        State state;
+        if (caseDataCopy.getHelpWithFeesAppliedForFees().toBoolean()) {
+            state = AwaitingHWFDecision;
```
nitinprabhu, line 81: do we have an integration to test AwaitingHWFDecision and  AwaitingPayment scenarios ?

mapingo: Should AwaitingHWFDecision.validate be called before moving to this state?

mapingo: Actually it looks like there is no validate for that state, so forget that

src/functionalTest/java/uk/gov/hmcts/divorce/citizen/CitizenSubmitApplicationTest.java
velimd: didnt really check the BE but in the FE we have a validation rule where the marriage must be atleast a year to get a divorce.

src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenSubmitApplication.java
velimd: Nitins Question: I did include a test and response.json file.

Shauns Question: yh there isnt, also i wasnt sure if i should add it as we have a event to validate this. Even AwaitingPayment.validate doesnt really validate anything relating to payment but checks casedata is valid.

mapingo (approved):

nitinprabhu (approved):

src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenSubmitApplication.java
```diff
@@ -81,9 +75,26 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
                 .state(Draft)
                 .build();
         }
+
+        State state;
+        if (caseDataCopy.getHelpWithFeesAppliedForFees() != null
```
nitinprabhu, line 80: if caseDataCopy.getHelpWithFeesAppliedForFees() is null what should be the default state ?

velimd: from the FE application stand, it should never be null, you have to answer that question. but here, included it, i can remove it as that field will be hit in the FE

nitinprabhu: from backend perspective it will be always set either to yes or no so should be never null

PR #323, opened 2021-06-03
fix: null pointer when HWF not entered

SebFlippence:

[NFDIV-680](https://tools.hmcts.net/jira/browse/NFDIV-680) Payment - State Changes

### Change description ###

Fixes a null pointer exception when HWF is not set.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenSubmitApplication.java
+++ b/src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenSubmitApplication.java
@@ -55,7 +55,7 @@ public class CitizenSubmitApplication implements CCDConfig<CaseData, State, User
 
     public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(CaseDetails<CaseData, State> details,
                                                                        CaseDetails<CaseData, State> beforeDetails) {
-        log.info("Submit application about to start callback invoked");
+        log.info("Submit application about to submit callback invoked");
 
         CaseData data = details.getData();
         CaseData caseDataCopy = data.toBuilder().build();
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenSubmitApplication.java
```diff
@@ -77,7 +77,7 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
         }
 
         State state;
-        if (caseDataCopy.getHelpWithFeesAppliedForFees().toBoolean()) {
+        if (caseDataCopy.getHelpWithFeesAppliedForFees() != null && caseDataCopy.getHelpWithFeesAppliedForFees().toBoolean()) {
```
nitinprabhu, line 80: can front end set this to No by default ?

velimd (approved):

src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenSubmitApplication.java
SebFlippence: `helpWithFeesNeedHelp` always gets set to `Yes`/`No` while `helpWithFeesAppliedForFees` only gets set once the first question is `Yes`. Perhaps we should be either checking if `helpWithFeesNeedHelp` 
 is true or if `helpWithFeesReferenceNumber` is not null and not empty

velimd:
```suggestion
        if (caseDataCopy.getHelpWithFeesNeedHelp().toBoolean() && caseDataCopy.getHelpWithFeesAppliedForFees().toBoolean()) {
```

SebFlippence: I've switched this to use `helpWithFeesNeedHelp` instead which will always be set.

nitinprabhu (approved):

velimd (approved):

PR #326, opened 2021-06-04
NFDIV-696

tomxelliott:
Set due date for applicant 2 to complete their part of the application

https://tools.hmcts.net/jira/browse/NFDIV-696

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/ccd/NoFaultDivorce.java
+++ b/src/main/java/uk/gov/hmcts/divorce/ccd/NoFaultDivorce.java
@@ -20,7 +20,7 @@ import static uk.gov.hmcts.divorce.common.model.access.Permissions.READ;
 @Component
 public class NoFaultDivorce implements CCDConfig<CaseData, State, UserRole> {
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/divorce/ccd/NoFaultDivorce.java
```diff
@@ -20,7 +20,7 @@
 @Component
 public class NoFaultDivorce implements CCDConfig<CaseData, State, UserRole> {
 
-    public static final String CASE_TYPE = "NO_FAULT_DIVORCE16";
+    public static final String CASE_TYPE = "NO_FAULT_DIVORCE17";
```
linusnorton, line 23: you don't need to bump the ID here, it's only when you make a backwards incompatible change like changing a field type

nitinprabhu, line 23: no need to bump as field type is not changed

linusnorton (approved):

nitinprabhu (approved):

PR #328, opened 2021-06-07
updating state

velimd:
### Change description ###
For the frontend submission page, there is a progression bar which informs the state of the case however, some of the state hasnt been implemented yet, just adding the state for FE check.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/ccd/NoFaultDivorce.java
+++ b/src/main/java/uk/gov/hmcts/divorce/ccd/NoFaultDivorce.java
@@ -20,7 +20,7 @@ import static uk.gov.hmcts.divorce.common.model.access.Permissions.READ;
 @Component
 public class NoFaultDivorce implements CCDConfig<CaseData, State, UserRole> {
 
-    public static final String CASE_TYPE = "NO_FAULT_DIVORCE16";
+    public static final String CASE_TYPE = "NO_FAULT_DIVORCE17";
     public static final String JURISDICTION = "DIVORCE";
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/divorce/ccd/NoFaultDivorce.java
```diff
@@ -20,7 +20,7 @@
 @Component
 public class NoFaultDivorce implements CCDConfig<CaseData, State, UserRole> {
 
-    public static final String CASE_TYPE = "NO_FAULT_DIVORCE16";
+    public static final String CASE_TYPE = "NO_FAULT_DIVORCE17";
```
tomxelliott, line 23: Not sure this needs updating. I thought it needed to be a backward incompatible change. Seeing as you are just adding new states should be fine. I may be wrong though 😬

nitinprabhu: This does not need updating if you are adding new state

src/main/java/uk/gov/hmcts/divorce/common/model/State.java
```diff
@@ -93,7 +93,10 @@
         name = "Application issued",
         label = "# **${[CASE_REFERENCE]}** ${applicant1LastName} **&** ${applicant2LastName}\n### **${[STATE]}**\n"
     )
-    Issued("Issued");
+    Issued("Issued"),
+
+    ConditionalOrderComplete("ConditionalOrderComplete"),
+    FinalOrderComplete("FinalOrderComplete");
```
nitinprabhu, line 99: there are no such states in divorce 1 so just thinking should we do it in another way ? @linusnorton would like to know your thoughts on this :)

nitinprabhu (approved):

PR #347, opened 2021-06-09
Fix wording fault on jurisdiction page

mapingo:
### Change description ###
Minor fix to wording on Jurisdiction page - replace apple with apply


nitinprabhu (approved):
🍎

PR #350, opened 2021-06-09
Solicitor model

linusnorton:
Re-apply the Solicitor model but exclude the OrganisationPolicy and Address fields when they are set to null. This will prevent CCD returning them as an empty object on the next call.

If we ever need to null these fields out I suggest we modify the `@JsonCreator` of those types to return null when they receive an empty object.


linusnorton:
@nitinprabhu you've already seen all this. The only bit you'll be interested in is: https://github.com/hmcts/nfdiv-case-api/pull/350/commits/fe6177561e667565d300a3715d60bde3b2561427

nitinprabhu (approved):

PR #351, opened 2021-06-10
content updates

linusnorton:


src/main/java/uk/gov/hmcts/divorce/solicitor/event/page/Applicant2ServiceDetails.java
```diff
@@ -51,7 +51,7 @@ public void addTo(final PageBuilder pageBuilder) {
                             .done()
                         .optional(OrganisationPolicy::getOrgPolicyCaseAssignedRole,
                             "applicant1NameChanged=\"NeverShow\"",
-                            APPLICANT_1_SOLICITOR)
+                            APPLICANT_2_SOLICITOR)
```
linusnorton, line 54: @mapingo / @nitinprabhu I think this was set to solicitor1 but I thought it should be 2 - is that right?

nitinprabhu: there is a never show condition on this field so just thinking can we just remove `APPLICANT_2_SOLICITOR`? and have ```.optional(OrganisationPolicy::getOrgPolicyCaseAssignedRole,
                            "applicant1NameChanged=\"NeverShow\"")```

linusnorton: I think there is a never show and a default value of APPLICANT_2_SOLICITOR if we take away the default value how will it get set?

nitinprabhu: ` APPLICANT_2_SOLICITOR("[APPTWOSOLICITOR]", "CRU")` is the role right rather than default value ?

linusnorton: it's the default value for the complex type to event json. 

```
  {
    "FieldDisplayOrder": 13,
    "DefaultValue": "[APPTWOSOLICITOR]",
    "LiveFrom": "01/01/2017",
    "CaseFieldID": "applicant2SolicitorOrganisationPolicy",
    "CaseEventID": "solicitor-create-application",
    "DisplayContext": "OPTIONAL",
    "ListElementCode": "OrgPolicyCaseAssignedRole",
    "FieldShowCondition": "applicant1NameChanged=\"NeverShow\""
  },

```

I'm fairly certain it needs to be applicant 2 rather than 1, otherwise it would set the wrong role

nitinprabhu (approved):

src/main/java/uk/gov/hmcts/divorce/solicitor/event/page/Applicant2ServiceDetails.java
nitinprabhu: got it now yeah it should be APPLICANT_2_SOLICITOR

tomxelliott (approved):

PR #352, opened 2021-06-10
NFDIV-784: Integration with Document Management Service.

SomeshRA:
NFDIV-784

### Change description ###
 Integration with Document Management Service.

```
[ ] Yes
[ x] No
```


src/main/java/uk/gov/hmcts/divorce/print/BulkPrintService.java
```diff
@@ -48,4 +62,21 @@ public UUID print(Print print) {
                 )))
             .letterId;
     }
+
+    private byte[] getDocumentBytes(final String docUrl, final String authToken)  {
+        final var userAuth = request.getHeader(AUTHORIZATION);
+        final var userDetails = idamService.retrieveUser(userAuth).getUserDetails();
+        var resourceResponseEntity = documentManagementClient.downloadBinary(
```
nitinprabhu, line 69: it is not clear what is the output type so we should not use var here

src/main/java/uk/gov/hmcts/divorce/print/BulkPrintService.java
```diff
@@ -25,19 +30,28 @@
 
     private final SendLetterApi sendLetterApi;
     private final AuthTokenGenerator authTokenGenerator;
+    private final HttpServletRequest request;
+    private final DocumentManagementClient documentManagementClient;
+    private final IdamService idamService;
```
nitinprabhu, line 35: normally we auto wired fields and also we don't have all args constructor on the service. So let's keep it consistent

src/main/java/uk/gov/hmcts/divorce/print/BulkPrintService.java
```diff
@@ -48,4 +62,21 @@ public UUID print(Print print) {
...
+            userAuth,
+            authToken,
+            String.join(",", userDetails.getRoles()),
+            userDetails.getId(),
+            FilenameUtils.getName(docUrl)
+        );
+        try {
+            return resourceResponseEntity.getBody().getInputStream().readAllBytes();
+        } catch (Exception e) {
+            throw new BulkPrintException("Doc url " + docUrl, e);
```
nitinprabhu, line 79: it's Document download exception rather than bulk print

src/main/java/uk/gov/hmcts/divorce/print/BulkPrintService.java
SomeshRA: Changed to specific type.

src/main/java/uk/gov/hmcts/divorce/print/BulkPrintService.java
SomeshRA: Added auto wired instead of all args  constructor.

src/main/java/uk/gov/hmcts/divorce/print/BulkPrintService.java
SomeshRA: Changed to DocumentDownload exception.

src/main/java/uk/gov/hmcts/divorce/common/config/advice/GlobalExceptionHandler.java
```diff
@@ -52,4 +53,11 @@
             String.format("%s - %s", exception.getMessage(), exception.contentUTF8())
         );
     }
+
+    @ExceptionHandler(DocumentDownloadException.class)
+    public ResponseEntity<Object> handleBadRequestException() {
```
nitinprabhu, line 58: method name seems incorrect

src/main/java/uk/gov/hmcts/divorce/common/config/advice/GlobalExceptionHandler.java
```diff
@@ -52,4 +53,11 @@
...
+        return new ResponseEntity<>(
+            HttpStatus.BAD_REQUEST
```
nitinprabhu, line 60: if the download fails it can be due to 403/401 or 5XX ?

SomeshRA: I suppose this would be handled by handleFeignException method.

src/main/java/uk/gov/hmcts/divorce/common/config/advice/GlobalExceptionHandler.java
SomeshRA: Any suggestions?

src/main/java/uk/gov/hmcts/divorce/common/config/advice/GlobalExceptionHandler.java
SomeshRA: Changed it to HttpStatus.INTERNAL_SERVER_ERROR.

src/main/java/uk/gov/hmcts/divorce/common/config/advice/GlobalExceptionHandler.java
SomeshRA: Changed it to - handleDocumentDownloadException

PR #394, opened 2021-06-24
Add caseworker note

nitinprabhu:

https://tools.hmcts.net/jira/browse/NFDIV-842

### Change description ###

- This PR adds event for caseworkers to add notes to case.

image

image

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
```diff
@@ -419,6 +421,22 @@
...
+    )
+    private List<ListValue<CaseNote>> notes;
+
+    @CCD(
+        label = "Add a case note",
+        hint = "Enter note",
+        typeOverride = TextArea,
+        access = {CaseworkerAndSuperUserAccess.class}
+    )
+    private String note;
```
linusnorton, line 438: Is it not possible just to add a complex type Note and the Note::getNote field to the event rather than have an extra field?

linusnorton (approved):

src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
nitinprabhu: discussed on slack

PR #400, opened 2021-06-28
Add new notes to the top of the  list and update list value ids

nitinprabhu:
image

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerAddNote.java
+++ b/src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerAddNote.java
@@ -16,12 +16,12 @@ import uk.gov.hmcts.divorce.common.model.UserRole;
 import uk.gov.hmcts.divorce.idam.IdamService;
 import uk.gov.hmcts.reform.idam.client.models.User;
 
-import javax.servlet.http.HttpServletRequest;
 import java.time.Clock;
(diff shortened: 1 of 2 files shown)


SomeshRA (approved):

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerAddNote.java
```diff
@@ -101,13 +102,16 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
 
             caseData.setNotes(listValues);
         } else {
+            AtomicInteger listValueIndex = new AtomicInteger(0);
```
linusnorton, line 105: Seems a bit OTT given this will be on the stack and not concurrent but I guess it doesn't hurt

linusnorton (approved):

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerAddNote.java
mapingo, line 105: Does this need to be an AtomicInteger?

nitinprabhu: doesn't need to be an atomic integer we can do it in traditional counter way but thought will use this as it will auto increment :)

PR #402, opened 2021-06-28
Update to changedNameHow field

lara-mpt:
https://tools.hmcts.net/jira/browse/NFDIV-753

### Change description ###
Update the changeNameHow field so that it can be used as checkboxes on the FE 

```
[ ] Yes
[x] No
```


nitinprabhu (commented):
you will need to bump the case type for this as the type of field has changed.

But also you will need to update gas type configured for manage cases in xui

lara-mpt:
> you will need to bump the case type for this as the type of field has changed.
> 
> But also you will need to update gas type configured for manage cases in xui

PR #405, opened 2021-06-28
Create HelpWithFees model

linusnorton:


src/main/java/uk/gov/hmcts/divorce/common/model/CaseData.java
```diff
@@ -88,31 +88,16 @@
...
-    private String helpWithFeesReferenceNumber;
-
-    @CCD(
-        label = "Need help with fees?",
-        access = {DefaultAccess.class}
-    )
-    private YesOrNo helpWithFeesNeedHelp;
+    @JsonUnwrapped(prefix = "helpWithFees")
+    @CCD(access = {DefaultAccess.class})
+    private HelpWithFees helpWithFees;
```
nitinprabhu, line 93: should this have ` @Builder.Default` and should we initialise this ?

linusnorton: I thought about it but it didn't seem necessary. There weren't any bits of code that assumed this would be set

nitinprabhu (approved):

PR #406, opened 2021-06-28
Move financial order to Applicant, relax tests

linusnorton:
Note this PR relaxes the integration and functional tests so they ignore extra fields


src/functionalTest/java/uk/gov/hmcts/divorce/caseworker/CaseworkerAddNoteTest.java
```diff
@@ -32,6 +33,7 @@ public void shouldUpdateCaseDataWithNotesWhenAboutToSubmitCallbackIsInvoked() th
         // notes.date value is compared using ${json-unit.any-string}
         // assertion will fail if the above value is missing
         assertThatJson(response.asString())
+            .when(IGNORING_EXTRA_FIELDS)
```
nitinprabhu, line 36: If you update/add new fields then expected response will still pass even if we forgot to add new fields to the response json ?

linusnorton: I think I'm okay with that. The json in the tests is testing for specific behaviour, the rest is just noise.

nitinprabhu (approved):

PR #415, opened 2021-06-30
NFDIV-965: Move AOS questions out of Applicant into AcknowledgementOf…

SomeshRA:

NFDIV-965

### Change description ###
Move AOS questions out of Applicant into AcknowledgementOfService and added aboutToSubmit call back 

```
[ ] Yes
[x ] No
```


linusnorton (approved):
We'll need more permissions later for the citizen side but this will do for now

PR #419, opened 2021-07-02
implementation of applicant2 confirmation of case info

velimd:
NFDIV-721

### Change description ###

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/divorce/common/model/Application.java
```diff
@@ -270,6 +270,18 @@
     @JsonFormat(pattern = "yyyy-MM-dd'T'HH:mm:ss.SSS")
     private LocalDateTime dateSubmitted;
 
+    @CCD(
+        label = "Is the information provided on the case correct?",
+        access = {DefaultAccess.class}
+    )
+    private YesOrNo applicant2ConfirmApplicant1Information;
```
nitinprabhu, line 277: do we have any event which will use these fields ? We can include fields in the test so we know they are covered.

velimd: im not sure, The ticket didnt mention the state change, i know there should be an event if they say no (was just mentioned in the review session) so i assume its another ticket. and if they say yes, they continue with their application.

nitinprabhu (approved):

PR #439, opened 2021-07-07
NFDIV-982

tomxelliott:
https://tools.hmcts.net/jira/browse/NFDIV-982


src/main/java/uk/gov/hmcts/divorce/common/model/access/Applicant2Access.java
```diff
@@ -0,0 +1,31 @@
...
+import static uk.gov.hmcts.divorce.common.model.UserRole.CASEWORKER_COURTADMIN_RDU;
+import static uk.gov.hmcts.divorce.common.model.UserRole.CASEWORKER_LEGAL_ADVISOR;
+import static uk.gov.hmcts.divorce.common.model.UserRole.CASEWORKER_SYSTEMUPDATE;
+import static uk.gov.hmcts.divorce.common.model.UserRole.CREATOR;
+
+public class Applicant2Access implements HasAccessControl {
+    @Override
+    public SetMultimap<HasRole, Permission> getGrants() {
+        SetMultimap<HasRole, Permission> grants = HashMultimap.create();
+        grants.putAll(CASEWORKER_SYSTEMUPDATE, Permissions.CREATE_READ_UPDATE);
```
nitinprabhu, line 21: should this be only read or this is used by bulkscan and hence need CRD ?

tomxelliott: I believe it can be R only. Just testing locally to make sure first - then will push the change

nitinprabhu (approved):

src/main/java/uk/gov/hmcts/divorce/common/model/Application.java
```diff
@@ -41,7 +42,7 @@
 
     @CCD(
         label = "Has applicant 2's marriage broken down irretrievably?",
-        access = {DefaultAccess.class}
+        access = {DefaultAccess.class, Applicant2Access.class}
```
linusnorton, line 45: I think you should be able to remove DefaultAccess here as the [CREATOR] will never write this field

linusnorton (approved):

PR #440, opened 2021-07-07
Make hidden fields read only to workaround deletion by exui

nitinprabhu:

https://tools.hmcts.net/jira/browse/NFDIV-984

### Change description ###

- Exui has existing bug where it deletes hidden fields https://tools.hmcts.net/jira/browse/EUI-3922
- There is a workaround at the moment where if we define hidden fields as read only then they are retained https://github.com/hmcts/fpl-ccd-configuration/pull/2809/files#diff-6aa4933bb605a7c8a13f35354cf5d77f9e94384ea705e24f4a93596c8a15f0adR33-R35

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorSubmitApplication.java
```diff
@@ -105,6 +105,7 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
         log.info("Submit application about to submit callback invoked");
 
         final CaseData caseData = details.getData();
+        log.info("APPLICANT 1 FINANCIAL ORDER BEFORE {} ", details.getData().getApplicant1().getFinancialOrder());
```
mapingo, line 108: Do we need to keep these log messages?

nitinprabhu: removed them

mapingo (approved):

linusnorton (approved):

PR #444, opened 2021-07-08
Add issue aos about to submit functional test

mapingo:
https://tools.hmcts.net/jira/browse/NFDIV-808

### Change description ###
Add issue aos event about to submit callback functional test
Fix flip flopping CitizenLinkApplicant2Test integration test


src/integrationTest/java/uk/gov/hmcts/divorce/citizen/event/CitizenLinkApplicant2Test.java
```diff
@@ -133,9 +131,6 @@ public void givenValidAccessCodeWhenCallbackIsInvokedThenAccessCodeIsRemovedAndS
         assertThatJson(actualResponse)
             .when(IGNORING_EXTRA_FIELDS)
             .isEqualTo(json(expectedCcdAboutToStartCallbackSuccessfulResponse()));
-
-        verify(serviceTokenGenerator).generate();
-        verifyNoMoreInteractions(serviceTokenGenerator);
```
nitinprabhu, line 138: is there a reason to remove this ?

src/functionalTest/java/uk/gov/hmcts/divorce/testutil/ServiceAuthenticationGenerator.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @Value("${idam.s2s-auth.url}")
+    private String s2sUrl;
+
+    public String generate() {
+        return generate(this.s2sName);
+    }
+
+    public String generate(final String s2sName) {
+        final Response response = RestAssured
```
nitinprabhu, line 30: there is one in https://github.com/hmcts/nfdiv-case-api/blob/master/src/functionalTest/java/uk/gov/hmcts/divorce/testutil/FunctionalTestSuite.java#L55 can we not use that ?

src/integrationTest/java/uk/gov/hmcts/divorce/citizen/event/CitizenLinkApplicant2Test.java
mapingo: This fails for me sometimes, and always fails on the PR, because we now have scheduled tasks that run at the same time as the test runs.  We can never know whether multiple calls to the ServiceTokenGenerator occur.

src/functionalTest/java/uk/gov/hmcts/divorce/testutil/ServiceAuthenticationGenerator.java
mapingo: I pinched this one from FunctionalTestSuite.java and the FunctionalTestSuite now uses this one, as I needed it to be more reusable across for other classes.

nitinprabhu (approved):

src/integrationTest/java/uk/gov/hmcts/divorce/citizen/event/CitizenLinkApplicant2Test.java
mapingo: I have taken a closer look at this problem and came up with a better solution

nitinprabhu: that's much neater nice one @mapingo

PR #456, opened 2021-07-12
Updated charts to use system user and fix integration test

nitinprabhu:


src/integrationTest/java/uk/gov/hmcts/divorce/solicitor/SolicitorSubmitApplicationTest.java
```diff
@@ -361,10 +361,6 @@ void givenCaseDataWithApplicationDocumentAndServiceAuthValidationFailsInDocStore
             .andExpect(
                 status().isUnauthorized()
             );
-
-        verify(serviceTokenGenerator).generate();
-
-        verifyNoMoreInteractions(serviceTokenGenerator);
```
nitinprabhu, line 367: this is now removed as we have scheduled tasks that run at the same time as the test runs.

linusnorton (approved):

PR #465, opened 2021-07-14
NFDIV-983: Ignoring case with null issue date

SomeshRA:
NFDIV-983

### Change description ###
Ignoring case with null issue date

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/caseworker/schedule/CaseworkerAwaitingConditionalOrderTask.java
+++ b/src/main/java/uk/gov/hmcts/divorce/caseworker/schedule/CaseworkerAwaitingConditionalOrderTask.java
@@ -52,7 +52,7 @@ public class CaseworkerAwaitingConditionalOrderTask {
                     Map<String, Object> caseDataMap = caseDetails.getData();
                     CaseData caseData = objectMapper.convertValue(caseDataMap, CaseData.class);
                     if (caseData.getIssueDate() == null) {
-                        log.info("Ignoring case id {} with created on {} and modified on {}, as issue date is null",
+                        log.error("Ignoring case id {} with created on {} and modified on {}, as issue date is null",
                             caseDetails.getId(),
                             caseDetails.getCreatedDate(),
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/divorce/caseworker/schedule/CaseworkerAwaitingConditionalOrderTask.java
```diff
@@ -51,13 +51,21 @@ public void execute() {
                 try {
                     Map<String, Object> caseDataMap = caseDetails.getData();
                     CaseData caseData = objectMapper.convertValue(caseDataMap, CaseData.class);
-                    long weeksBetweenIssueDateAndNow = WEEKS.between(caseData.getIssueDate(), LocalDate.now());
-
-                    if (weeksBetweenIssueDateAndNow > HOLDING_PERIOD_IN_WEEKS) {
-                        log.info("Case id {} has been in holding state for > 20 weeks hence moving state to AwaitingConditionalOrder",
-                            caseDetails.getId()
+                    if (caseData.getIssueDate() == null) {
```
nitinprabhu, line 54: could you please invert this and put this in else part
if(issue!=null){}
else{log...}

nitinprabhu: Also it should log as error rather than info

SomeshRA: PMD rule failing with : 
Avoid negation within an "if" expression with an "else" clause. For example, rephrase: if (x != y) diff(); else same(); as: if (x == y) same(); else diff();.

SomeshRA: Changed logging to error.

nitinprabhu (approved):

PR #509, opened 2021-07-23
Fix for change in idam env vars

linusnorton:
Fix for https://github.com/hmcts/ccd-data-store-api/commit/7aaea598ec77a756df825631d157ab41080b32b1


nitinprabhu (approved):

charts/nfdiv-case-api/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: "1.0"
 description: A Helm chart for nfdiv-case-api App
 name: nfdiv-case-api
 home: https://github.com/hmcts/nfdiv-case-api
-version: 0.0.29
+version: 0.0.30
```
nitinprabhu, line 6: Not needed if we update preview yaml

PR #512, opened 2021-07-23
Nfdiv 988

tomxelliott:
https://tools.hmcts.net/jira/browse/NFDIV-988


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Application.java
```diff
@@ -219,21 +219,42 @@
...
     )
-    private YesOrNo legalProceedings;
+    private YesOrNo applicant1LegalProceedings;
 
     @CCD(
         label = "What do the legal proceedings relate to?",
         access = {DefaultAccess.class}
     )
-    private Set<LegalProceedingsRelated> legalProceedingsRelated;
+    private Set<LegalProceedingsRelated> applicant1LegalProceedingsRelated;
```
velimd, line 228: would it not be best to move these to the applicant files?

velimd (commented):
github is saying there are 3 empty json files, is there a reason for this?

linusnorton (commented):
Just a suggestion but you could make a LegalProceedings class with both of those fields then inside Application add two properties:
```
@JsonUnwrapped(prefix = "applicant1")
private LegalProceedings applicant1;
@JsonUnwrapped(prefix = "applicant2")
private LegalProceedings applicant2;
```

linusnorton (approved):

PR #518, opened 2021-07-26
Update to changeNameHow field

lara-mpt:
https://tools.hmcts.net/jira/browse/NFDIV-753

### Change description ###
Update the changeNameHow field so that it can be used as checkboxes on the FE

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/divorcecase/NoFaultDivorce.java
+++ b/src/main/java/uk/gov/hmcts/divorce/divorcecase/NoFaultDivorce.java
@@ -20,7 +20,7 @@ import static uk.gov.hmcts.divorce.divorcecase.model.access.Permissions.READ;
 @Component
 public class NoFaultDivorce implements CCDConfig<CaseData, State, UserRole> {
 
-    public static final String CASE_TYPE = "NO_FAULT_DIVORCE18";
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/divorce/divorcecase/NoFaultDivorce.java
```diff
@@ -20,7 +20,7 @@
 @Component
 public class NoFaultDivorce implements CCDConfig<CaseData, State, UserRole> {
 
-    public static final String CASE_TYPE = "NO_FAULT_DIVORCE18";
+    public static final String CASE_TYPE = "NFD";
```
nitinprabhu, line 23: should be NO_FAULT_DIVORCE 🤔

lara-mpt: I thought we were changing it now to NFD? - https://github.com/hmcts/cnp-flux-config/pull/10235/files

linusnorton: yes it's `NFD`

linusnorton (approved):

PR #536, opened 2021-07-29
use latest java chart

linusnorton:


charts/nfdiv-case-api/Chart.yaml
```diff
@@ -3,12 +3,12 @@ appVersion: "1.0"
 description: A Helm chart for nfdiv-case-api App
 name: nfdiv-case-api
 home: https://github.com/hmcts/nfdiv-case-api
-version: 0.0.30
+version: 0.0.31
```
nitinprabhu, line 6: chart version doesn't need bumping

nitinprabhu (approved):

PR #547, opened 2021-07-30
NFDIV-770

tomxelliott:
Set up event to withdraw case when switched to sole application

https://tools.hmcts.net/jira/browse/NFDIV-770


tomxelliott:
Closed as ticket withdrawn

PR #602, opened 2021-08-16
App1 resubmit notification

velimd:
https://tools.hmcts.net/jira/browse/NFDIV-989
https://tools.hmcts.net/jira/browse/NFDIV-990

### Change description ###

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenApplicant1Resubmit.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Override
+    public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuilder) {
+
+        configBuilder
+            .event(APPLICANT_1_RESUBMIT)
+            .forStates(AwaitingApplicant1Response)
+            .name("Applicant 1 Resubmit")
+            .description("Applicant 1 resubmits for joint application")
+            .grant(CREATE_READ_UPDATE, CITIZEN)
```
linusnorton, line 41: shouldn't this be APP ONE?

linusnorton: no I guess that would give it to solicitors

linusnorton (approved):
I think we need to add an applicant1 role but we'll have to sort that later

PR #604, opened 2021-08-16
nfdiv-724 add switched to sole event and state

lara-mpt:
https://tools.hmcts.net/jira/browse/NFDIV-724

### Change description ###
add the switched to sole event and state

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenSwitchedToSole.java
```diff
@@ -0,0 +1,51 @@
...
+                                                                       CaseDetails<CaseData, State> beforeDetails) {
+        log.info("Applicant 1 switched to sole about to submit callback invoked");
+
+        CaseData data = details.getData();
+
+        data.setApplicationType(ApplicationType.SOLE_APPLICATION);
+
+        return AboutToStartOrSubmitResponse.<CaseData, State>builder()
+            .data(data)
+            .state(SwitchedToSole)
```
nitinprabhu, line 48: ccd will move the state as you defined it here https://github.com/hmcts/nfdiv-case-api/pull/604/files#diff-0f746b2fd3c36667a15b9b736875a002e298206a6837524a6ca7ffd791e32551R30 so we won't need to explicitly set the state

lara-mpt: @nitinprabhu did you mean that the line 48 `.state(SwitchedToSole)` was unnecessary,  or are you referencing line 44 
`data.setApplicationType(ApplicationType.SOLE_APPLICATION);` which is explicitly setting the application type not the state

nitinprabhu (approved):

PR #623, opened 2021-08-19
PBA Account number retrieval and payments using payments v2 API

nitinprabhu:

https://tools.hmcts.net/jira/browse/NFDIV-546

### Change description ###

- Retrieval of PBA numbers for a given solicitor account.
- Making PBA Payment using Payment API v2.
- Error handling for payment failures.
- Added/Updated unit, integration and functional tests.
- Removed unwanted method in solicitor submit application.

### Pending
- Add some more integration and functional tests
- This PR needs to be merged after docker setup is available for PBA account retrieval and payments.

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/payment/PaymentService.java
+++ b/src/main/java/uk/gov/hmcts/divorce/payment/PaymentService.java
@@ -24,6 +24,7 @@ import uk.gov.hmcts.reform.authorisation.generators.AuthTokenGenerator;
 import javax.servlet.http.HttpServletRequest;
 import java.io.IOException;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 
 import static java.util.Collections.singletonList;
@@ -40,9 +41,9 @@ public class PaymentService {
     private static final String FAMILY = "family";
     private static final String FAMILY_COURT = "family court";
     private static final String DIVORCE = "divorce";
-    public static final String GBP = "GBP";
-    private static final String CAE0001 = "CA-E0001";
-    private static final String CAE0004 = "CA-E0004";
+    private static final String GBP = "GBP";
+    public static final String CAE0001 = "CA-E0001";
+    public static final String CAE0004 = "CA-E0004";
 
     @Autowired
     private HttpServletRequest httpServletRequest;
@@ -73,29 +74,8 @@ public class PaymentService {
             .build();
     }
 
-    private ListValue<Fee> getFee(final FeeResponse feeResponse) {
-        return ListValue
-            .<Fee>builder()
-            .value(
-                Fee
-                    .builder()
-                    .amount(getValueInPence(feeResponse.getAmount()))
-                    .code(feeResponse.getFeeCode())
-                    .description(feeResponse.getDescription())
-                    .version(String.valueOf(feeResponse.getVersion()))
-                    .build()
-            )
-            .build();
-    }
-
     public PbaResponse processPbaPayment(CaseData caseData, Long caseId) {
         log.info("Processing PBA payment for case id {}", caseId);
-
-        feesAndPaymentsClient.creditAccountPayment(
-            httpServletRequest.getHeader(AUTHORIZATION),
-            authTokenGenerator.generate(),
-            creditAccountPaymentRequest(caseData, caseId)
-        );
 
         ResponseEntity<CreditAccountPaymentResponse> paymentResponseResponseEntity = null;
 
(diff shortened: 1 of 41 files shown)


src/main/java/uk/gov/hmcts/divorce/payment/PaymentService.java
```diff
@@ -39,6 +80,63 @@ public OrderSummary getOrderSummary() {
             .build();
     }
 
+    public PbaResponse processPbaPayment(CaseData caseData, Long caseId) {
```
linusnorton, line 83: It would be nice is this didn't take the whole case data. Is it possible to just pass the OrderSummary for the thing that we're trying to take payment for? I'm thinking that in future we might have other things to pay for like amends, in which case we'll need to send a different OrderSummary

src/main/java/uk/gov/hmcts/divorce/payment/PaymentService.java
```diff
@@ -53,4 +151,121 @@ public OrderSummary getOrderSummary() {
...
+    private CreditAccountPaymentRequest creditAccountPaymentRequest(CaseData caseData, Long caseId) {
+        var creditAccountPaymentRequest = new CreditAccountPaymentRequest();
+        var orderSummary = caseData.getApplication().getApplicationFeeOrderSummary();
+
+        creditAccountPaymentRequest.setService(DIVORCE_SERVICE);
+        creditAccountPaymentRequest.setCurrency(GBP);
+        creditAccountPaymentRequest.setSiteId(caseData.getSelectedDivorceCentreSiteId());
+        creditAccountPaymentRequest.setAccountNumber(getPbaNumber(caseData));
+
+        final var solicitor = caseData.getApplicant1().getSolicitor();
```
linusnorton, line 222: As far as the design of this class goes I think the public method needs to take two things: what the OrderSummary is (what are we paying for) and who is paying for it (app1/app2 solicitor). It's probably not good to hard code to app1 solicitor even if it's all we need now

src/main/java/uk/gov/hmcts/divorce/payment/PaymentService.java
nitinprabhu: payment request requires data which is in case data https://github.com/hmcts/nfdiv-case-api/pull/623/files#diff-86d92efa4d934715621e6205346aa501f609e65f6665b0ccdd3b6dad23de70faR213-R239
 so we will need to pass case data anyways even if it is amend or create.

src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorSubmitApplication.java
```diff
@@ -139,6 +130,24 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
...
         }
 
+        final var application = caseData.getApplication();
+        final var applicationFeeOrderSummary = application.getApplicationFeeOrderSummary();
+
+        if (caseData.getApplication().isSolicitorPaymentMethodPba()) {
+            PbaResponse response = paymentService.processPbaPayment(caseData, caseId);
+
+            if (response.getHttpStatus() == CREATED) {
+                updateCaseDataWithPaymentDetails(applicationFeeOrderSummary,caseData, response.getPaymentReference());
```
linusnorton, line 140:
```suggestion
                updateCaseDataWithPaymentDetails(applicationFeeOrderSummary, caseData, response.getPaymentReference());
```

src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorSubmitApplication.java
```diff
@@ -139,6 +130,24 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
...
+            } else {
+                return AboutToStartOrSubmitResponse.<CaseData, State>builder()
+                    .data(details.getData())
+                    .errors(singletonList(response.getErrorMessage()))
+                    .build();
+            }
+        }
```
linusnorton, line 147: Shouldn't there be an else statement here? I guess the only other option is HWF?

nitinprabhu: yes we have only two options so it's either paid via PBA or HWF

src/main/java/uk/gov/hmcts/divorce/payment/PaymentService.java
nitinprabhu: makes sense will pass solicitor as an argument to payment service method

src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorSubmitApplication.java
linusnorton: Ah, I guess that happens in the case submission tasks

mapingo (approved):

linusnorton (approved):

PR #715, opened 2021-09-20
NFDIV-239: Set assigned role depending on the application type

tomxelliott:
https://tools.hmcts.net/jira/browse/NFDIV-239


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/access/AosAccess.java
```diff
@@ -21,6 +22,7 @@
         grants.putAll(LEGAL_ADVISOR, Permissions.READ);
         grants.putAll(APPLICANT_2_SOLICITOR, Permissions.CREATE_READ_UPDATE);
         grants.putAll(APPLICANT_2, Permissions.CREATE_READ_UPDATE);
```
linusnorton, line 24: Maybe applicant2 role can be removed?

tomxelliott: Yeah I am not sure why Somesh added it originally

PR #716, opened 2021-09-20
Reissue application

nitinprabhu:

https://tools.hmcts.net/jira/browse/NFDIV-865

### Change description ###

- Includes Reissue application event with callback to handle 3 types of reissue i.e Digital Aos, Offline Aos and complete reissue case.
- Documents generated during issue and reissue will now have date time appended to filename.
- Includes unit and integration test.
- Functional tests will be implemented in a separate PR.


src/integrationTest/java/uk/gov/hmcts/divorce/caseworker/CaseworkerIssueApplicationIT.java
```diff
@@ -167,7 +168,7 @@ static void tearDown() {
     @BeforeEach
     void setClock() {
         when(clock.instant()).thenReturn(Instant.parse("2021-06-17T12:00:00.000Z"));
-        when(clock.getZone()).thenReturn(ZoneId.systemDefault());
+        when(clock.getZone()).thenReturn(ZoneId.of("Europe/London"));
```
linusnorton, line 171: I don't think this is correct - servers run on UTC

linusnorton: unless it's for the document filename.... which I guess it is

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerReissueApplication.java
```diff
@@ -0,0 +1,95 @@
...
+            .aboutToSubmitCallback(this::aboutToSubmit)
+            .grant(CREATE_READ_UPDATE, CASE_WORKER)
+            .grant(READ,
+                SOLICITOR,
+                SUPER_USER,
+                LEGAL_ADVISOR))
+            .page("reissueApplication")
+            .pageLabel("Reissue Divorce Application")
+            .complex(CaseData::getApplication)
+            .mandatory(Application::getReissueOption)
```
linusnorton, line 67: formatting :)

src/main/java/uk/gov/hmcts/divorce/caseworker/service/ReIssueApplicationService.java
```diff
@@ -0,0 +1,126 @@
...
+    private Integer daysUntilOverdue;
+
+    public CaseDetails<CaseData, State> process(final CaseDetails<CaseData, State> caseDetails) {
+        ReissueOption reissueOption = caseDetails.getData().getApplication().getReissueOption();
+
+        log.info("For case id {} reissue option selected is {} ", caseDetails.getId(), reissueOption.getLabel());
+
+        caseDetails.getData().getApplication().setReissueDate(now(clock));
+
+        CaseDetails<CaseData, State> updatedCaseDetails = null;
```
linusnorton, line 62: it's a minor point, but when I see code like this (a switch statement that sets a variable) I think it's slightly cleaner to move the switch statement into a method and then use return values, so this would be:

```
var updatedCaseDetails = updateCase(reissueOption, details);
```

src/main/java/uk/gov/hmcts/divorce/caseworker/service/ReIssueApplicationService.java
```diff
@@ -0,0 +1,126 @@
...
+                        details.getData().setDueDate(now(clock).plusDays(daysUntilOverdue));
+                        return details;
+                    },
+                    setPostIssueState,
+                    sendApplicationIssueNotifications
+                ).run(caseDetails);
+                break;
+
+            default:
+                log.info("For case id {} invalid reissue option hence not processing reissue application ", caseDetails.getId());
```
linusnorton, line 116: I know this should never happen but you will get a null pointer error here because updatedCaseDetails is null and you use it below.

linusnorton (approved):

PR #720, opened 2021-09-21
Fix sonar quality gate issue

nitinprabhu:


src/main/java/uk/gov/hmcts/divorce/caseworker/service/ReIssueApplicationService.java
```diff
@@ -55,12 +57,16 @@
...
 
-        log.info("For case id {} reissue option selected is {} ", caseDetails.getId(), reissueOption.getLabel());
-
-        caseDetails.getData().getApplication().setReissueDate(now(clock));
+        log.info("For case id {} reissue option selected is {} ", caseDetails.getId(), reissueOption);
 
         var updatedCaseDetails = updateCase(caseDetails, reissueOption);
 
+        if (isNull(updatedCaseDetails)) {
+            throw new ReissueProcessingException(
```
mapingo, line 65: It might be cleaner to make updateCase impossible to return a null, and instead add this exception as the default for the switch statement

nitinprabhu: updated code as discussed

mapingo (approved):

PR #723, opened 2021-09-22
Set Europe/London timezone for the app

nitinprabhu:
### Change description ###

- By default UTC timezone is used due to which documents generated during issue/reissue has time stamp which is one hour behind.
- Set Europe/London as default timezone for the app.
- Tested on local docker and is working fine.
image


mark-butler-solirius (approved):
lgtm

mapingo (approved):

PR #752, opened 2021-09-29
NFDIV-1127: Handle conditional content required for joint applications

tomxelliott:


src/main/java/uk/gov/hmcts/divorce/citizen/notification/conditionalorder/Applicant1ApplyForConditionalOrderNotification.java
```diff
@@ -16,6 +16,10 @@
 @Slf4j
 public class Applicant1ApplyForConditionalOrderNotification {
 
+    public static final String JOINT_CONDITIONAL_ORDER = "Your partner has also received this notification."
```
velimd, line 19: if this is a notificiation template variable, i think it might be best to reduce this and use a short name. as its just going to be replaced.

velimd (approved):

PR #775, opened 2021-10-04
Add Statement of Truth page to submit conditional order event

mapingo:
https://tools.hmcts.net/jira/browse/NFDIV-1120

### Change description ###
Add Statement of Truth page to submit conditional order event


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/ConditionalOrder.java
```diff
@@ -70,4 +71,18 @@
...
 
+    @CCD(
+        label = "Solicitor’s name"
+    )
+    private String sotName;
+
+    @CCD(
+        label = "Solicitor’s firm"
+    )
+    private String sotFirm;
```
nitinprabhu, line 82: should be solicitorFirm rather than statement of truth firm ? same goes for other fields

nitinprabhu (approved):

PR #796, opened 2021-10-11
NFDIV-1252: Refine ES queries so that no redundant results are returned

tomxelliott:
https://tools.hmcts.net/jira/browse/NFDIV-1252


src/main/java/uk/gov/hmcts/divorce/systemupdate/schedule/SystemProgressCasesToAosOverdueTask.java
```diff
@@ -49,7 +49,8 @@ public void run() {
         final String serviceAuth = authTokenGenerator.generate();
 
         try {
-            final List<CaseDetails> casesInAwaitingAosState = ccdSearchService.searchForAllCasesWithStateOf(AwaitingAos, user, serviceAuth);
+            final List<CaseDetails> casesInAwaitingAosState =
+                ccdSearchService.searchForAllCasesWithStateOf(AwaitingAos, null, user, serviceAuth);
```
mapingo, line 53: Maybe there should be two search methods, one that has the flag parameter and one without.  That way we don't have to pass the null as an argument, internally the methods can call a single implementation.

src/main/java/uk/gov/hmcts/divorce/systemupdate/service/CcdSearchService.java
```diff
@@ -62,15 +65,44 @@ public SearchResult searchForCaseWithStateOf(final State state,
             sourceBuilder.toString());
     }
 
-    public List<CaseDetails> searchForAllCasesWithStateOf(final State state, User user, String serviceAuth) {
+    public SearchResult searchForCasesWithStateOfDueDateBeforeWithoutFlagSet(final State state,
+                                                                             final int from,
+                                                                             final int size,
+                                                                             final String flag,
```
nitinprabhu, line 71: would be nice to call this as searchQueryParam or something similar

tomxelliott: Sorry what do you mean?

nitinprabhu: I mean instead of flag we could call it searchQueryParam :)

nitinprabhu (approved):

PR #801, opened 2021-10-12
Nfdiv 1228 bug app2 not represented

mark-butler-solirius:

https://tools.hmcts.net/jira/browse/NFDIV-1228

### Change description ###

Solicitor2Represented flag was incorrectly set to Yes during SolicitorSubmitApplication if the Solicitor object in Applicant2 was not null. 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorSubmitApplication.java
```diff
@@ -199,11 +199,11 @@ private void updateCaseDataWithPaymentDetails(
     }
 
     private void updateApplicant2DigitalDetails(CaseData caseData) {
-        if (caseData.getApplicant2().getSolicitor() != null) {
+        if (caseData.getApplicant2().getSolicitor() != null
+            && caseData.getApplicant2().getSolicitor().getOrganisationPolicy() != null) {
```
nitinprabhu, line 203: If the solicitor is not null that means on XUI user has selected app2 is represented by solicitor.
Also if user selects it is represented by solicitor then organisation policy has to be entered else it will error on UI.
which scenario you had where solicitor object was not null and org policy was null ?

nitinprabhu: discussed on slack

PR #840, opened 2021-10-21
Implement error handling for bulk case creation

nitinprabhu:
### Change description ###

- If updating bulk list case reference fails for awaiting pronouncement cases then those cases would now be removed from bulk case as otherwise it will keep appearing in other batches.

### DIFF
--- a/config/owasp/suppressions.xml
+++ b/config/owasp/suppressions.xml
@@ -20,5 +20,6 @@
       <notes>Tomcat 9.0.45</notes>
       <cve>CVE-2021-30640</cve>
       <cve>CVE-2021-33037</cve>
+      <cve>CVE-2021-42340</cve>
     </suppress>
 </suppressions>


src/main/java/uk/gov/hmcts/divorce/bulkaction/ccd/event/RemoveFailedCaseLinksFromBulkCase.java
```diff
@@ -0,0 +1,32 @@
...
+
+import static uk.gov.hmcts.divorce.bulkaction.ccd.BulkActionState.Created;
+import static uk.gov.hmcts.divorce.divorcecase.model.UserRole.CASE_WORKER;
+import static uk.gov.hmcts.divorce.divorcecase.model.UserRole.SYSTEMUPDATE;
+import static uk.gov.hmcts.divorce.divorcecase.model.access.Permissions.CREATE_READ_UPDATE;
+
+@Component
+@Slf4j
+public class RemoveFailedCaseLinksFromBulkCase implements CCDConfig<BulkActionCaseData, BulkActionState, UserRole> {
+    public static final String REMOVE_FAILED_CASE_LINKS = "remove-failed-case-links";
```
linusnorton, line 20: Maybe `system-remove-failed-cases`

PR #879, opened 2021-11-01
NFDIV-1275 & NFDIV-1288: Remove cases from Bulk Case

tomxelliott:
https://tools.hmcts.net/jira/browse/NFDIV-1288


linusnorton (approved):
I think the only thing I don't like is the ValidationUtil but that's not your doing

nitinprabhu (approved):

PR #882, opened 2021-11-01
Nfdiv 1270 add cron process failed pronounced cases

mapingo:
https://tools.hmcts.net/jira/browse/NFDIV-1270

### Change description ###
Add SystemRetryBulkCasePronouncedErrors cron job


src/main/java/uk/gov/hmcts/divorce/bulkaction/service/CasePronouncementService.java
```diff
@@ -37,6 +36,9 @@
     @Autowired
     private IdamService idamService;
 
+    @Autowired
+    private ScheduleCaseService scheduleCaseService;
```
nitinprabhu, line 40: I think this should be renamed otherwise casepronoucement service is composed of schedule case service.
May be we should rename ScheduleCaseService to BulkCaseProcessingService  or something similar ?

mapingo: I have taken a look, and I think it will be best to move the CaseTask Factory method out into its own service. I have also found some duplicated code in ScheduleCaseService, I will see if that can be refactored away too.

mapingo: updated in:
https://github.com/hmcts/nfdiv-case-api/pull/882/commits/c36a0042b87aaf977e3699ea73184a298f0ed8e1
https://github.com/hmcts/nfdiv-case-api/pull/882/commits/1a58835f3d7abaacc2ee9fc11b43a80fc8734f3c

PR #901, opened 2021-11-05
Share a case

linusnorton:
- Remove [CREATOR] role when adding [APPONESOLICITOR] role
- Switch roles on submitted hook in solicitor-create-application event instead of aboutToStart of solicitor-submit-case
- Upgrade ccd-client library
- Use Case Assignment API to change user roles to ensure supplementary data for Share a case is set correctly by CCD

As noted on: https://tools.hmcts.net/confluence/display/RCCD/A+Guide+to+Assign+Access+to+cases+for+professional+users%3A+configuration?focusedCommentId=1475839441#comment-1475839441 you must remove the [CREATOR] role before adding the new role or the supplementary data does not get set.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -246,7 +246,7 @@ dependencies {
   implementation group: 'uk.gov.hmcts.reform', name: 'logging-spring', version: versions.reformsJavaLogging
 
   implementation group: 'uk.gov.hmcts.reform', name: 'idam-client', version: '2.0.0'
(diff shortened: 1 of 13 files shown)


linusnorton:
See also: https://github.com/hmcts/cnp-flux-config/pull/12525/files

nitinprabhu (approved):

PR #939, opened 2021-11-17
Nfdiv 1425 mulitple bulk case creation

mapingo:
https://tools.hmcts.net/jira/browse/NFDIV-1425

### Change description ###
Add multiple bulk case creation for each cron job run
Only process cases once so that failures are not added to multiple cases
If all cases fail to update then the bulk case will be set to the Empty state

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/bulkaction/ccd/event/SystemEmptyCase.java
+++ b/src/main/java/uk/gov/hmcts/divorce/bulkaction/ccd/event/SystemEmptyCase.java
@@ -14,7 +14,7 @@ import static uk.gov.hmcts.divorce.divorcecase.model.UserRole.SYSTEMUPDATE;
 import static uk.gov.hmcts.divorce.divorcecase.model.access.Permissions.CREATE_READ_UPDATE;
 
 @Component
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/hmcts/divorce/bulkaction/ccd/event/SystemEmptyCase.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public static final String SYSTEM_EMPTY_CASE = "system-empty-case";
+
+    @Override
+    public void configure(final ConfigBuilder<BulkActionCaseData, BulkActionState, UserRole> configBuilder) {
+        new BulkActionPageBuilder(configBuilder
+            .event(SYSTEM_EMPTY_CASE)
+            .forStateTransition(Created, Empty)
+            .name("Set empty case")
+            .description("Set empty case")
```
tomxelliott, line 27: Minor thing really but perhaps we could have a slightly more descriptive description for anyone who doesn't really know what the empty case functionality is for or what it does?

tomxelliott: Might not be necessary though since it is merely a state transition event

tomxelliott (approved):

PR #943, opened 2021-11-18
NFDIV-1348: Add checks for whether app2 email is known

tomxelliott:
https://tools.hmcts.net/jira/browse/NFDIV-1348

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/caseworker/service/task/SendApplicationIssueNotifications.java
+++ b/src/main/java/uk/gov/hmcts/divorce/caseworker/service/task/SendApplicationIssueNotifications.java
@@ -37,7 +37,8 @@ public class SendApplicationIssueNotifications implements CaseTask {
                 }
             } else {
                 applicationIssuedNotification.sendToJointApplicant1(caseData, caseId);
-                if (caseData.getApplication().getApplicant1KnowsApplicant2EmailAddress().toBoolean()) {
+                if (Objects.nonNull(caseData.getApplication().getApplicant1KnowsApplicant2EmailAddress())
+                    && caseData.getApplication().getApplicant1KnowsApplicant2EmailAddress().toBoolean()) {
                     applicationIssuedNotification.sendToJointApplicant2(caseData, caseId);
                 }
             }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/divorce/caseworker/service/task/SendApplicationIssueNotifications.java
```diff
@@ -37,7 +37,9 @@
                 }
             } else {
                 applicationIssuedNotification.sendToJointApplicant1(caseData, caseId);
-                applicationIssuedNotification.sendToJointApplicant2(caseData, caseId);
+                if (caseData.getApplication().getApplicant1KnowsApplicant2EmailAddress().toBoolean()) {
+                    applicationIssuedNotification.sendToJointApplicant2(caseData, caseId);
```
mapingo, line 41: Could getApplicant1KnowsApplicant2EmailAddress() ever return null?

tomxelliott: To get to this point they would have had to ask this question, but it never hurts to be defensive with our programming so will add a check 👍

mapingo (approved):

src/main/java/uk/gov/hmcts/divorce/caseworker/service/task/SendApplicationIssueNotifications.java
```diff
@@ -37,7 +37,10 @@
                 }
             } else {
                 applicationIssuedNotification.sendToJointApplicant1(caseData, caseId);
-                applicationIssuedNotification.sendToJointApplicant2(caseData, caseId);
+                if (Objects.nonNull(caseData.getApplication().getApplicant1KnowsApplicant2EmailAddress())
```
nitinprabhu, line 40: do we need a test to ensure when this is null `applicationIssuedNotification.sendToJointApplicant2` is not executed ?

tomxelliott: @nitinprabhu - yes I did try that but due to the fact that the same mock is used for the other email execution I couldn't use verifyNoInteractions

PR #957, opened 2021-11-22
Nfdiv 1352 service application tab formatting

mark-butler-solirius:

https://tools.hmcts.net/jira/browse/NFDIV-1352

### Change description ###
Change the archival of the ServiceApplication to create a new Object (ServiceApplicationOutcome). This new object will be displayed for the "Previous Service Applications" and will support the new DisplayOrder field so that the NFD formatting resembles the Divorce1 formatting 

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/AlternativeServiceOutcome.java
```diff
@@ -0,0 +1,134 @@
...
+    private YesOrNo successfulServedByBailiff;
+
+    @CCD(
+        label = "Reason for failure to serve",
+        typeOverride = TextArea,
+        displayOrder = "15"
+    )
+    private String reasonFailureToServeByBailiff;
+
+    public String getServiceApplicationOutcomeLabel() {
```
linusnorton, line 131: does this need an @JsonIgnore or should it be included?

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/AlternativeServiceOutcome.java
```diff
@@ -0,0 +1,134 @@
...
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+@Builder
+public class AlternativeServiceOutcome {
+
+    @CCD(
+        label = "Application date",
+        displayOrder = "1"
```
linusnorton, line 26: should these be integers?

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/CaseData.java
```diff
@@ -364,4 +366,24 @@ public void archiveAlternativeServiceApplicationOnCompletion() {
         }
     }
 
+    @SuppressWarnings("PMD")
+    public AlternativeServiceOutcome buildAlternativeServiceOutcome(AlternativeService alternativeService) {
```
linusnorton, line 370: it looks to me like this method should be on AlternativeService something like alternativeService.getOutcome()

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/AlternativeServiceOutcome.java
mark-butler-solirius: Yeah. I initially created the displayOrder attribute as an Integer in the CCD Interface class, but I couldn't initially work out how to get the default value in the Interface to be null. I'm happy to go back and change it, or come up with another solution if we cannot use null defaults.

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/CaseData.java
mark-butler-solirius: Yep - will move it

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/AlternativeServiceOutcome.java
linusnorton: assuming it defaults to 0 we could treat that is null

mark-butler-solirius: Ok - will implement this solution

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/AlternativeService.java
```diff
@@ -108,4 +109,25 @@
     @CCD(access = {CaseworkerAccessBetaOnlyAccess.class})
     private Bailiff bailiff = new Bailiff();
 
+    @SuppressWarnings("PMD")
+    @JsonIgnore
+    public AlternativeServiceOutcome getOutcome() {
+        return AlternativeServiceOutcome.builder()
+            .alternativeServiceType(this.getAlternativeServiceType())
```
linusnorton, line 116: is there any need to use getters here? maybe just use the local property

mark-butler-solirius: Yes, I can change to the local properties (didn't think about it after I moved this method from the CaseData class)

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/AlternativeServiceOutcome.java
mark-butler-solirius: Yes - this should be included so the Label gets added to the AlternativeServiceOutcome list object

linusnorton (approved):

PR #993, opened 2021-11-29
Fix multiselect field type and removed enum value which is not required

nitinprabhu:


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/ConditionalOrder.java
```diff
@@ -141,7 +142,7 @@
         typeOverride = MultiSelectList,
         typeParameterOverride = "ClarificationReason"
     )
-    private ClarificationReason refusalClarificationReason;
+    private Set<ClarificationReason> refusalClarificationReason;
```
linusnorton, line 145: won't this break things?

linusnorton (approved):

PR #1006, opened 2021-11-30
NFDIV-1454: Set up email for clarification submitted

tomxelliott:
https://tools.hmcts.net/jira/browse/NFDIV-1454

Screenshot 2021-11-30 at 10 11 36


src/main/java/uk/gov/hmcts/divorce/legaladvisor/event/LegalAdvisorMakeDecision.java
```diff
@@ -47,6 +52,7 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
             .showSummary()
             .showEventNotes()
             .aboutToSubmitCallback(this::aboutToSubmit)
+            .submittedCallback(this::submitted)
```
nitinprabhu, line 55: is there a reason notification is sent in submitted can we not send notification as part of about to submit so in case notification fails legal advisor can retrigger it ?

src/main/java/uk/gov/hmcts/divorce/legaladvisor/event/LegalAdvisorMakeDecision.java
```diff
@@ -108,4 +114,18 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
...
+
+    public SubmittedCallbackResponse submitted(final CaseDetails<CaseData, State> details,
+                                               final CaseDetails<CaseData, State> beforeDetails) {
+
+        final CaseData caseData = details.getData();
+        final State state = details.getState();
+
+        if (caseData.getApplication().isSolicitorApplication()
+            && AwaitingClarification.equals(state)) {
+            notification.send(caseData, details.getId());
```
nitinprabhu, line 126: adding a log would be nice including case id and which notification is sent

src/main/java/uk/gov/hmcts/divorce/legaladvisor/event/LegalAdvisorMakeDecision.java
tomxelliott: Happy to move to about to submit - just thought having it there would ensure the state had been correctly set in about to submit

src/main/java/uk/gov/hmcts/divorce/legaladvisor/event/LegalAdvisorMakeDecision.java
tomxelliott: Cool, will add

nitinprabhu (approved):

PR #1032, opened 2021-12-06
remove bulk case creation duplicated code

mapingo:
### Change description ###
Refactor SystemCreateBulkCaseListTask to use BulkTriggerService and remove duplicated code
BulkTriggerService getCaseTask now has the Bulk Case Details as an argument


nitinprabhu (approved):
nice one Shaun

PR #1072, opened 2021-12-13
Use different field for citizen update case state

lara-mpt:
### Change description ###
Use different field for citizen update case state
The previous field applicant2SolicitorAddress was only accessible for applicant 1 users

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenUpdateCaseStateAat.java
+++ b/src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenUpdateCaseStateAat.java
@@ -43,8 +43,8 @@ public class CitizenUpdateCaseStateAat implements CCDConfig<CaseData, State, Use
 
         CaseData data = details.getData();
 
-        State state = State.valueOf(data.getApplicant2().getSolicitor().getAddress());
-        data.getApplicant2().getSolicitor().setAddress(null);
+        State state = State.valueOf(data.getApplicant2().getMiddleName());
+        data.getApplicant2().setMiddleName(null);
 
         return AboutToStartOrSubmitResponse.<CaseData, State>builder()
             .data(data)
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenUpdateCaseStateAat.java
```diff
@@ -43,8 +43,8 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
 
         CaseData data = details.getData();
 
-        State state = State.valueOf(data.getApplicant2().getSolicitor().getAddress());
-        data.getApplicant2().getSolicitor().setAddress(null);
+        State state = State.valueOf(data.getApplicant2().getMiddleName());
+        data.getApplicant2().setMiddleName(null);
```
nitinprabhu, line 47: for unit test you can just set ` caseDetails.setState(State.Holding);` instead of using data in the case.
Currently it is looking like you are setting middle state from middle name

nitinprabhu: In your test you can do 
```
 caseDetails.setState(State.Holding);
        final AboutToStartOrSubmitResponse<CaseData, State> response = citizenUpdateCaseStateAat.aboutToSubmit(caseDetails, caseDetails);

```

lara-mpt: @nitinprabhu the input from that middle name field should be what's setting the state of the case. As that's how we'll update the state from one request

lara-mpt: so that's why the unit test is testing that the state changes if the middle name field is set

nitinprabhu: ok got you

nitinprabhu (approved):

PR #1089, opened 2021-12-14
Content changes for financial order page

nitinprabhu:
https://tools.hmcts.net/jira/browse/NFDIV-1494
image


linusnorton (approved):
previous field was financialOrderFor so it shouldn't have any issues

PR #1105, opened 2021-12-16
Nfdiv 1495 pay account content update

mark-butler-solirius:

https://tools.hmcts.net/jira/browse/NFDIV-1495

### Change description ###
Update Pay on Account content

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/CaseData.java
```diff
@@ -38,6 +38,7 @@
 import static uk.gov.hmcts.ccd.sdk.type.FieldType.FixedRadioList;
 import static uk.gov.hmcts.ccd.sdk.type.FieldType.TextArea;
 
+@SuppressWarnings("PMD")
```
nitinprabhu, line 41: do we need this suppression ?

nitinprabhu (approved):

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/CaseData.java
mark-butler-solirius: Good spot. I added it when I was having trouble with the ccd import and a static constant in this file. But I changed how the suppression method works. I'll remove it

PR #1115, opened 2021-12-16
NFDIV-1204: Add new fields to Conditional Order model

tomxelliott:
https://tools.hmcts.net/jira/browse/NFDIV-1204


nitinprabhu (approved):

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Applicant.java
```diff
@@ -168,6 +168,18 @@
     )
     private YesOrNo continueApplication;
 
+    @CCD(
+        label = "Is the information in this application still correct?"
+    )
+    private YesOrNo confirmInformationStillCorrect;
```
linusnorton, line 174: I would put these questions in the section they come up in, like ConditionalOrder or something

tomxelliott: Is it not a similar situation to the one the other day where both applicants will be answering the question or is it just an app1 question @linusnorton ?

linusnorton (approved):

PR #1121, opened 2021-12-17
NFDIV-1569: Move conditional order questions into their own class

tomxelliott:
https://tools.hmcts.net/jira/browse/NFDIV-1569


nitinprabhu (commented):
Do we need to migrate old fields ?

tomxelliott:
The only 2 that do not have the same json value anymore are 2 fields that have not been fully integrated on the citizen side so no migration required. One new field that I added the other day was removed and has also not been integrated on the frontend as well.

nitinprabhu (approved):

linusnorton (approved):

PR #1124, opened 2021-12-17
Add notification interface and dispatcher

mapingo:
https://tools.hmcts.net/jira/browse/NFDIV-1568


src/main/java/uk/gov/hmcts/divorce/common/service/task/SendCitizenSubmissionNotifications.java
```diff
@@ -24,43 +25,24 @@
...
     @Override
     public CaseDetails<CaseData, State> apply(final CaseDetails<CaseData, State> caseDetails) {
 
         final CaseData caseData = caseDetails.getData();
         final Long caseId = caseDetails.getId();
         final State state = caseDetails.getState();
 
-        if (!caseData.getApplication().isSolicitorApplication()) {
-            sendCitizenNotifications(caseData, caseId, state);
+        if (Submitted == state || AwaitingDocuments == state || AwaitingHWFDecision == state) {
```
nitinprabhu, line 38: can do
 ```
if(EnumSet.of(AwaitingDocuments,Submitted,AwaitingHWFDecision).contains(details.getState())){

        }
```

File: src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Application.java [hunk 2]
```diff
@@ -467,4 +467,15 @@ public boolean isSolicitorApplication() {
     public boolean isSolicitorPaymentMethodPba() {
         return FEE_PAY_BY_ACCOUNT.equals(this.getSolPaymentHowToPay());
     }
+
+    @JsonIgnore
+    public boolean isApplicant1OffLine() {
```
linusnorton, line 472: shouldn't these be Applicant.isOffline?

linusnorton (approved):

nitinprabhu (approved):

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Application.java
mapingo: Yes, that is what I wanted, but at the moment the only way to work out whether to do off line or not is "applicant1KnowsApplicant2EmailAddress" field in Application.  Moving that field inside the Applicant class, would not work.  We could have a new field inside Applicant that is set on submission, that way we can set the appropriate offline setting for applicant1 and applicant2.

PR #1251, opened 2022-01-25
nfdiv-1157 update conditional order tab

lara-mpt:
https://tools.hmcts.net/jira/browse/NFDIV-1157

### Change description ###
update conditional order tab for joint citizen application

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/divorce/divorcecase/tab/CaseTypeTab.java
```diff
@@ -200,10 +200,22 @@ private void buildConditionalOrderTab(ConfigBuilder<CaseData, State, UserRole> c
...
             .field("coApplicant1ChangeOrAddToApplication")
             .field("coApplicant1StatementOfTruth")
+            .label("labelConditionalOrderDetails-Applicant2",
+                "applicationType=\"jointApplication\" AND coApplicant2ApplyForConditionalOrder=\"*\"",
+                "### Applicant 2")
+            .field("coApplicant2ApplyForConditionalOrder")
+            .field("coApplicant1ConfirmInformationStillCorrect")
+            .field("coApplicant2SubmittedDate")
+            .field("coApplicant2ChangeOrAddToApplication")
+            .field("coApplicant2StatementOfTruth")
```
nitinprabhu, line 218: do you want to display these fields only for joint ?

lara-mpt: @nitinprabhu yes but I thought they'd only show up if they were filled in, and they'd only be filled in in a joint application. 
I can add the condition for them if you prefer?

Just noticed coApplicant1ConfirmInformationStillCorrect should be coApplicant2ConfirmInformationStillCorrect so I'll update that

nitinprabhu: normally they would be displayed only if they have some value. So if for sole none of the above field have values populated then it should be good otherwise we will need the condition.

nitinprabhu (approved):

PR #1286, opened 2022-01-31
Add full CCD error to the logs

linusnorton:


nitinprabhu (approved):

src/main/java/uk/gov/hmcts/divorce/systemupdate/service/CcdUpdateService.java
```diff
@@ -59,6 +59,7 @@ public void submitEvent(final CaseDetails caseDetails,
 
             final String message = format("Submit Event Failed for Case ID: %s, Event ID: %s", caseId, eventId);
             log.info(message, e);
+            log.info(e.contentUTF8());
```
nitinprabhu, line 62: will it disclose any case data in case of error ?

linusnorton: only if CCD does

nitinprabhu: safer side we can do only on non prod envs so we don't accidentally leak data ?

linusnorton: I would assume anything coming back from their API is going in their logs anyway. I wouldn't be comfortable showing the error to the user but I think putting it in the logs would be okay. I've never seen CCD return case data in an error message

PR #1296, opened 2022-02-01
remove label

linusnorton:
Labels are now optional. Where possible we should use labels defined in the event. The `@CCD` is also optional if no other options are required.


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/CaseData.java
```diff
@@ -243,13 +240,9 @@
     )
     private List<ListValue<ScannedDocument>> scannedDocuments;
 
-    @CCD(
-        label = "Supplementary evidence handled"
```
nitinprabhu, line 247: this is used 
image

nitinprabhu (approved):

PR #1417, opened 2022-02-24
adding final order fields for respondent

velimd:
https://tools.hmcts.net/jira/browse/NFDIV-1640

### Change description ###

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/divorcecase/model/FinalOrder.java
+++ b/src/main/java/uk/gov/hmcts/divorce/divorcecase/model/FinalOrder.java
@@ -11,6 +11,7 @@ import lombok.NoArgsConstructor;
 import uk.gov.hmcts.ccd.sdk.api.CCD;
 import uk.gov.hmcts.ccd.sdk.api.HasLabel;
 import uk.gov.hmcts.ccd.sdk.type.YesOrNo;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/FinalOrder.java
```diff
@@ -80,7 +81,19 @@
         label = "Does the applicant want to apply for Final Order and ${labelContentFinaliseDivorceOrEndCivilPartnership}?",
         access = {DefaultAccess.class}
     )
-    private YesOrNo doesApplicantWantToApplyForFinalOrder;
```
nitinprabhu, line 83: we will need migration in Retired Field class

velimd: if that is the case, i can just leave it as applicant. what do you think is better as a Back-end expert?

nitinprabhu: I think changes in your PR is better as we would be able to identify exactly it's applicant1 or app2.

nitinprabhu (approved):

PR #1432, opened 2022-02-28
clean up sol details in sole document

velimd:
https://tools.hmcts.net/jira/browse/NFDIV-1656

### Change description ###
[37a43945-9530-4f84-9d54-a8083bf46e9c.pdf](https://github.com/hmcts/nfdiv-case-api/files/8155251/37a43945-9530-4f84-9d54-a8083bf46e9c.pdf)

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/divorce/document/content/DivorceApplicationSoleTemplateContent.java
```diff
@@ -142,13 +141,29 @@ private void setSolicitorDetails(Map<String, Object> templateContent, Applicant
...
+        if (!isNullOrEmpty(solicitorEmail)) {
+            templateContent.put(APPLICANT_2_SOLICITOR_EMAIL, solicitorEmail);
+        }
+        if (!isNullOrEmpty(solicitorFirmName)) {
+            templateContent.put(APPLICANT_2_SOLICITOR_FIRM_NAME, solicitorFirmName);
+        }
+        if (!isNullOrEmpty(solicitorAddress)) {
+            String addressCleanUp =
+                join("\n",
+                    Arrays.stream(applicant.getSolicitor().getAddress().split("\n"))
```
velimd, line 163: the FE has different inputs for solicitor address so if the user only inputs line 1 and postcode for example, itll end up being `line1\n\n\n\npostcode` which doesnt not look great on the document hence the idea was to get rid of the extra new lines and have it look like `line1\npostcode`

mapingo: I was wondering if we need to do the same for the address for the applicants, as that may return the Solicitor address if they are represented

mapingo (approved):

src/main/java/uk/gov/hmcts/divorce/document/content/DivorceApplicationSoleTemplateContent.java
velimd: or unless a CW just added multiple lines in the textarea

PR #1446, opened 2022-03-01
Change applicant homeAddress to address

mapingo:
### Change description ###
Change and migrate applicant homeAddress to address


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Applicant.java
```diff
@@ -83,8 +83,8 @@
     )
     private String nameChangedHowOtherDetails;
 
-    @CCD(label = "Home address")
-    private AddressGlobalUK homeAddress;
+    @CCD(label = "Aaddress")
```
linusnorton, line 86: Should be `Address`

linusnorton (approved):

PR #1489, opened 2022-03-07
Add new util class to get all notify email templates that match a given search criteria

amoriartyMOJ:
https://tools.hmcts.net/jira/browse/NFDIV-1677

### Change description ###
With this util class we can now search a templates values for a dynamic content variable
or we can search the body of the template for static text.

Useful for when changes are needed to notify email templates we can now find which templates
from the list will need changing.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/notification/GetAllTemplatesFromNotify.java
+++ b/src/main/java/uk/gov/hmcts/divorce/notification/GetAllTemplatesFromNotify.java
@@ -5,8 +5,6 @@ import uk.gov.service.notify.NotificationClientException;
 import uk.gov.service.notify.Template;
 import uk.gov.service.notify.TemplateList;
 
-import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
--- a/src/main/java/uk/gov/hmcts/divorce/notification/GetAllTemplatesFromNotify.java
+++ b/src/main/java/uk/gov/hmcts/divorce/notification/GetAllTemplatesFromNotify.java
@@ -1,0 +1,103 @@
+package uk.gov.hmcts.divorce.notification;
+
+import uk.gov.service.notify.NotificationClient;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/hmcts/divorce/notification/GetAllTemplatesFromNotify.java
```diff
@@ -0,0 +1,103 @@
...
+        int count = 0;
+
+        for (Template template: templates.getTemplates()) {
+
+            Map<String, Object> personalisation = template.getPersonalisation().orElse(new HashMap<>());
+
+            if (personalisation.containsKey(searchCriteria)) {
+                System.out.println("************************************************************");
+
+                if (template.getName().contains("COR") || template.getName().contains("NOT USED") || template.getName().contains("OLD")) {
```
linusnorton, line 58: what's COR?

src/main/java/uk/gov/hmcts/divorce/notification/GetAllTemplatesFromNotify.java
```diff
@@ -0,0 +1,103 @@
...
+        int count = 0;
+
+        for (Template template: templates.getTemplates()) {
+
+            String body = template.getBody();
+
+            if (body.contains(searchCriteria)) {
+                System.out.println("************************************************************");
+
+                if (template.getName().contains("COR") || template.getName().contains("NOT USED") || template.getName().contains("OLD")) {
```
linusnorton, line 81: if this list is re-used maybe it should be `List.of(....)` in static var

linusnorton (approved):

src/main/java/uk/gov/hmcts/divorce/notification/GetAllTemplatesFromNotify.java
amoriartyMOJ: Leftover from original code, my mistake. Removed and extracted the if statement code to a common func

PR #1509, opened 2022-03-09
Integrate cftlib gradle plugin

banderous:
This integrates the latest cftlib which is now a gradle plugin. 

The plugin now handles some of the build setup and introduces a way of running automated tests with the new cftlibTest sourceset and test task.

I've added a placeholder test for now that will check the application loads alongside ccd.


nitinprabhu (approved):

build.gradle
```diff
@@ -407,24 +399,21 @@ ccd {
...
-        def key = it.substring(0, index)
-        def value = it.substring(index + 1)
-        environment key, value
-      }
+bootWithCCD {
+    dependsOn(loadEnvSecrets, buildCCDXlsx)
+    environment 'IDAM_S2S-AUTH_URL', '${IDAM_S2S_URL}'
+    environment 'CCD_DOCUMENT_URL_PATTERN', '.*'
+    environment 'ROLE_ASSIGNMENT_S2S_AUTHORISED_SERVICES', 'ccd_gw,am_role_assignment_service,am_org_role_mapping_service,xui_webapp,aac_manage_case_assignment,ccd_data,nfdiv_case_api'
+    environment 'DATA_STORE_S2S_AUTHORISED_SERVICES', 'ccd_gw,fpl_case_service,ccd_data,ccd_ps,divorce_frontend,payment-api,xui_webapp,nfdiv_case_api,ccd_case_document_am_api,am_role_assignment_service'
```
linusnorton, line 409: I wonder if we could move this from here to `RunWithCft` or `CftLibConfig` - just override these env vars for the services that need them rather than globally?

src/cftlib/java/uk/gov/hmcts/divorce/cftlib/RunWithCFT.java
```diff
@@ -1,19 +0,0 @@
...
-import uk.gov.hmcts.rse.ccd.lib.api.LibRunner;
-
-import java.util.Map;
-import java.util.Set;
-
-@SuppressWarnings("HideUtilityClassConstructor") // Spring needs a constructor, its not a utility class
-public class RunWithCFT {
-
-    public static void main(final String[] args) {
-        LibRunner.run(CaseApiApplication.class, Set.of(), Map.of(
```
linusnorton, line 13: I didn't mind having this class as it made it a bit more obvious how we could tweak and override bits for specific CCD apps

build.gradle
banderous: Yes service specific properties will be possible in future. We can probably come up with a cleaner API for defining these things rather than having to set these service-specific magic properties.

src/cftlib/java/uk/gov/hmcts/divorce/cftlib/RunWithCFT.java
banderous: The LibRunner is gone because we have a new bootstrapping process. These were always global overrides rather than service-specific however. Can add service-specific overrides in future.

linusnorton (approved):

PR #1574, opened 2022-03-16
NFDIV-1899 Shutter for solicitors in prod

linusnorton:


src/main/java/uk/gov/hmcts/divorce/divorcecase/NoFaultDivorce.java
```diff
@@ -28,5 +29,9 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
 
         // to shutter the service within xui uncomment this line
         // configBuilder.shutterService();
+
+        if (System.getenv().getOrDefault("ENVIRONMENT", "").equalsIgnoreCase("PROD")) {
```
nitinprabhu, line 33: I think we are not yet setting environment variable on prod

linusnorton: We use it for AAT in other places: https://github.com/hmcts/nfdiv-case-api/blob/master/src/main/java/uk/gov/hmcts/divorce/common/AddSystemUpdateRole.java

nitinprabhu: Yeah we set via Jenkins file there I think

PR #1581, opened 2022-03-16
NFDIV-1278 - Notification when the service application is successful

danielbe11:

https://tools.hmcts.net/jira/browse/NFDIV-1278

### Change description ###

- Sending notification when alternative service is granted.

```
[ ] Yes
[X] No
```


nitinprabhu (approved):

src/main/java/uk/gov/hmcts/divorce/legaladvisor/event/LegalAdvisorMakeServiceDecision.java
```diff
@@ -107,7 +111,9 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
         log.info("Application end state is {}", endState);
 
         serviceApplication.setServiceApplicationDecisionDate(LocalDate.now(clock));
-        if (serviceApplication.getServiceApplicationGranted().toBoolean()) {
+
+        boolean isServiceApplicationGranted = serviceApplication.getServiceApplicationGranted().toBoolean();
```
benouaer, line 115: same comment as above about possible null value
also, is the variable needed if it's only used in the if clause?

benouaer (approved):

src/main/java/uk/gov/hmcts/divorce/legaladvisor/event/LegalAdvisorMakeServiceDecision.java
danielbe11: Since we have .mandatory(AlternativeService::getServiceApplicationGranted), it can only ever be Yes or No so will leave the if / else condition as is.

PR #1610, opened 2022-03-21
NFDIV-1886 - Removing jurisdiction data for switch-to-sole

danielbe11:

https://tools.hmcts.net/jira/browse/NFDIV-1886

### Change description ###

- Removing jurisdiction data when switching to sole.

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenSwitchedToSole.java
```diff
@@ -129,4 +130,18 @@ private CaseData removeApplicant2AnswersFromCase(CaseData caseData) {
...
+        caseData.getApplication().getJurisdiction().setResidualEligible(null);
+        caseData.getApplication().getJurisdiction().setConnections(null);
+        caseData.getApplication().getJurisdiction().setApplicant1Residence(null);
+        caseData.getApplication().getJurisdiction().setApplicant2Residence(null);
+        caseData.getApplication().getJurisdiction().setBothLastHabituallyResident(null);
+        caseData.getApplication().getJurisdiction().setApp1HabituallyResLastTwelveMonths(null);
+        caseData.getApplication().getJurisdiction().setApp1HabituallyResLastSixMonths(null);
+
+        return caseData;
+    }
```
lara-mpt, line 146: I think you can just do 
```suggestion
    private CaseData removeJurisdictionAnswersFromCase(CaseData caseData) {
        caseData.getApplication().setJurisdiction(null);
        return caseData;
    }
```

lara-mpt: In which case you can just add it to the removeApplicant2AnswersFromCase()

src/test/java/uk/gov/hmcts/divorce/citizen/event/CitizenSwitchedToSoleTest.java
```diff
@@ -352,6 +354,72 @@ void shouldNotUnlinkApp2OnAboutToSubmitCallback() {
...
+                    AddressGlobalUK.builder()
+                        .addressLine1("123 The Street")
+                        .postTown("The town")
+                        .county("County Durham")
+                        .country("England")
+                        .postCode("POSTCODE")
+                        .build())
+                .contactDetailsType(PRIVATE)
+                .build()
+        );
```
lara-mpt, line 382: I don't think this is necessary since we're only testing the jurisdiction fields here

lara-mpt (approved):

PR #1668, opened 2022-03-29
NFDIV-2002

tomxelliott:
Make confirm receipt accessible to caseworker for offline cases

https://tools.hmcts.net/jira/browse/NFDIV-2002


src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorConfirmReceipt.java
```diff
@@ -18,19 +18,19 @@
 
 @Component
 public class SolicitorConfirmReceipt implements CCDConfig<CaseData, State, UserRole> {
```
nitinprabhu, line 20: worth renaming the class and move it to common package

nitinprabhu (approved):

PR #1672, opened 2022-03-29
NFDIV-1925 : NOP with confidential info

sunilnatikar:
### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1925

### Change description ###
Changes done to add NOP documents to Confidential Documents tab under section called "Confidential Documents Generated" when Caseworker issues an application and applicant contact is private. Updated bulk print logic to pick the documents accordingly.
On update contact detail events documents are moved between Confidential and non confidential documents tabs based on applicants contact details privacy flag.

**Updated areas:**
- Issue application event 
    - Document generation
    - Bulk print
- Contact details update events for caseworker and solicitor

```
[] Yes
[x] No
```


src/main/java/uk/gov/hmcts/divorce/document/print/model/Letter.java
```diff
@@ -1,14 +1,29 @@
...
 import uk.gov.hmcts.ccd.sdk.type.ScannedDocument;
+import uk.gov.hmcts.divorce.document.model.ConfidentialDivorceDocument;
 import uk.gov.hmcts.divorce.document.model.DivorceDocument;
 
-@AllArgsConstructor
 @Getter
 public class Letter {
-    private final DivorceDocument divorceDocument;
-    private final ScannedDocument scannedDocument;
+    private DivorceDocument divorceDocument;
```
linusnorton, line 10: why did you get rid of the `final` stuff?

sunilnatikar: as we have now added multiple constructor overloads with required fields only, so we don't have to set the optional fields as final otherwise we will have to set them as null in each constructor.

linusnorton (approved):

PR #1784, opened 2022-04-14
Feature/nfdiv 1865  Applicant 1 Solicitor answer received notification

benouaer:
### Change description ###
- implement Applicant 1 Solicitor answer received notification

- https://tools.hmcts.net/jira/browse/NFDIV-1865


src/main/java/uk/gov/hmcts/divorce/notification/CommonContent.java
```diff
@@ -114,18 +114,6 @@ public String getPartner(CaseData caseData, Applicant partner) {
         }
     }
 
-    public String getPartnersSolicitor(CaseData caseData, Applicant partner) {
```
velimd, line 117: this seems like itll be useful, why remove it?

velimd (approved):

src/main/java/uk/gov/hmcts/divorce/notification/CommonContent.java
benouaer: because it doesn't do much, it's the same as getPartner, just adds " 'solicitor" basically. so we can just use the other method
also, it was redoing all the logic that the getPartner method was doing

PR #1848, opened 2022-04-28
Access control changes

linusnorton:
### Change description ###

Revert the explicit app2 permissions added to the CAA classes and upgrade the ccd-config-generator so that the permissions are inherited from the parent class


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Solicitor.java
```diff
@@ -34,47 +32,42 @@ public class Solicitor {
 
     @CCD(
         label = "Solicitor’s name",
-        access = {DefaultAccess.class, Applicant2Access.class, CaseworkerWithCAAAccess.class}
```
nitinprabhu, line 37: Applicant has only default access ```    @CCD(access = {DefaultAccess.class})
    private Applicant applicant1 = new Applicant();``` so should we keep `Applicant2Access` ?

nitinprabhu (approved):

PR #1902, opened 2022-05-15
nfdiv-2189: retain currentState post state transition

benouaer:
### Change description ###

Enter a description.

https://tools.hmcts.net/jira/browse/NFDIV-2189

- [X] tests have been updated / new tests has been added (if needed)


src/main/java/uk/gov/hmcts/divorce/systemupdate/event/SystemAttachScannedDocuments.java
```diff
@@ -34,4 +39,16 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
                 .mandatoryWithLabel(BulkScanMetaInfo::getEvidenceHandled, "Supplementary evidence handled")
                 .done();
     }
+
+    public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(
+        final CaseDetails<CaseData, State> details,
+        final CaseDetails<CaseData, State> beforeDetails) {
+        log.info("[aboutToSubmit]: saving currentState({}) into caseData to retain it post state transition", details.getState());
+        var caseData = details.getData();
+        caseData.getApplication().setPreviousState(details.getState());
```
nitinprabhu, line 48: there is state transition `.forStateTransition(POST_SUBMISSION_STATES, OfflineDocumentReceived)` so not sure why we need  this callback. Also this will impact other scenarios where we expect state to be `OfflineDocumentReceived` whenever we received scanned documents

benouaer: As you can see from the code, the case state itself isn't affected. This is just keeping a record of what the state was before the transition. This is needed because in the frontend, when the state is OfflineDocumentReceived. In that scenarios, the hub pages and progressBar should remain as they were for the previousState, and should not be changed

nitinprabhu (approved):

PR #1943, opened 2022-05-23
NFDIV-1266: Add Solicitor Certificate of Entitlement Email notifications

tomxelliott:
https://tools.hmcts.net/jira/browse/NFDIV-1266


src/main/java/uk/gov/hmcts/divorce/citizen/notification/conditionalorder/EntitlementGrantedConditionalOrderNotification.java
```diff
@@ -46,6 +58,17 @@ public void sendToApplicant1(final CaseData caseData, final Long id) {
             caseData.getApplicant1().getLanguagePreference());
     }
 
+    @Override
+    public void sendToApplicant1Solicitor(final CaseData caseData, final Long id) {
+        log.info("Sending entitlement granted on conditional order notification to applicant 1 solicitor for case : {}", id);
+
+        notificationService.sendEmail(
```
nitinprabhu, line 65: ticket mentions this should be only done for digital applications ? Do we need to check if the case was created via bulk scan and filter those out ?

nitinprabhu (approved):

PR #1975, opened 2022-05-31
NFDIV-1298 : Conditional Order Granted doc generation

sunilnatikar:
### Change description ###
Changes done to generated Conditional Order Granted document when System-Pronounce-Case event is triggered.

### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-1298

- [x] this ticket been reviewed by QA
- [x] the user story been signed off by the PO


src/main/java/uk/gov/hmcts/divorce/systemupdate/service/task/GenerateConditionalOrderPronouncedDocument.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Override
+    public CaseDetails<CaseData, State> apply(CaseDetails<CaseData, State> caseDetails) {
+
+        final Long caseId = caseDetails.getId();
+        final CaseData caseData = caseDetails.getData();
+
+        log.info("Generating Conditional Order pronounced pdf for CaseID: {}", caseDetails.getId());
+
+        caseDataDocumentService.renderDocumentAndUpdateCaseData(
```
nitinprabhu, line 40: I think we will need to check if document was already generated and added to the case because if the bulk case event fails and caseworker retries this then we will add multiple documents?

sunilnatikar: @nitinprabhu reverted the doc store removal logic.

nitinprabhu (approved):

PR #2057, opened 2022-06-22
Set up CCD diff tool

linusnorton:
### Change description ###

Set up CCD diff tool


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/CaseData.java
```diff
@@ -78,7 +77,7 @@ public class CaseData {
 
     @JsonUnwrapped(prefix = "applicant2")
     @Builder.Default
-    @CCD(access = {DefaultAccess.class, Applicant2Access.class})
+    @CCD(access = {DefaultAccess.class})
```
nitinprabhu, line 80: once this is removed happy to merge

PR #2132, opened 2022-07-12
nfdiv-1444 welsh translation review state

lara-mpt:
### Change description ###

Add Welsh translation review state and set cases to it on certain events.

https://tools.hmcts.net/jira/browse/NFDIV-1444


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/State.java
```diff
@@ -342,25 +342,14 @@ public enum State {
...
-        AosOverdue,
-        AosDrafted,
-        AwaitingServicePayment,
-        AwaitingServiceConsideration,
-        AwaitingBailiffReferral,
-        AwaitingBailiffService,
-        IssuedToBailiff,
-        Holding,
-        AwaitingConditionalOrder
-    );
```
lara-mpt, line 363: removed this as it is unused

src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenFinalOrderDelayReason.java
```diff
@@ -51,14 +52,22 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(CaseDetails<C
...
 
         CaseData data = details.getData();
+        State endState = FinalOrderRequested;
 
         if (data.getApplicationType().isSole()) {
             notificationDispatcher.send(soleAppliedForFinalOrderNotification, data, details.getId());
         }
 
+        if (data.isWelshApplication()) {
+            data.getApplication().setWelshPreviousState(endState);
```
nitinprabhu, line 62: why not set the state from details.getState rather than hardcoding ? We need to set the existing state right ?

src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenFinalOrderDelayReason.java
```diff
@@ -51,14 +52,22 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(CaseDetails<C
...
+            endState = WelshTranslationReview;
+            log.info("State set to WelshTranslationReview, WelshPreviousState set to {}, CaseID {}",
+                data.getApplication().getWelshPreviousState(), details.getId());
+        }
+
         return AboutToStartOrSubmitResponse.<CaseData, State>builder()
             .data(data)
-            .state(FinalOrderRequested)
+            .state(endState)
```
nitinprabhu, line 70: this can then be set as `WelshTranslationReview`

src/main/java/uk/gov/hmcts/divorce/common/event/ApplyForFinalOrder.java
```diff
@@ -83,6 +84,13 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(CaseDetails<C
             }
 
             endState = FinalOrderRequested;
+
+            if (data.isWelshApplication()) {
+                data.getApplication().setWelshPreviousState(endState);
```
nitinprabhu, line 89: same here ?

src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenFinalOrderDelayReason.java
lara-mpt: @nitinprabhu we need to set the state that it would have become if it wasn't for the Welsh translation. 
In this case if we used details.getState() we'd get FinalOrderOverdue which is wrong as we should be setting it as FinalOrderRequested (i.e the state that this event would be set to if it wasn't for Welsh)

src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenFinalOrderDelayReason.java
lara-mpt: changing the state here would set it as WelshTranslationReview for all cases on this event regardless of the WelshLanguage fields. Although we could split this up into using two separate returns depending on which state to set it to

nitinprabhu: thanks for clarifying that makes sense

nitinprabhu (approved):

PR #2147, opened 2022-07-15
Nfdiv-2525 Add validation to submit aos event

lara-mpt:
### Change description ###

Add validation to submit aos event

https://tools.hmcts.net/jira/browse/NFDIV-2525


tomxelliott:
Only other field I can see causing an issue with Null Pointer Exceptions is `caseData.getApplicant2().getLegalProceedings().getValue()` on line 71 in `RespondentAnswersTemplateContent`. It is mandatory on the BE, not sure if it is also mandatory on the FE so need to ensure a value is passed otherwise we are going to keep getting NPE's on this line.

nitinprabhu (approved):

PR #2368, opened 2022-09-15
Nfdiv 2492 add migration for aos is drafted

mapingo:
### Change description ###

Add set aosIsDrafted to Yes system migration task

https://tools.hmcts.net/jira/browse/NFDIV-2829


src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerReissueApplication.java
```diff
@@ -61,7 +62,8 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
             .grantHistoryOnly(
                 SOLICITOR,
                 SUPER_USER,
-                LEGAL_ADVISOR))
+                LEGAL_ADVISOR,
+                SYSTEMUPDATE))
```
nitinprabhu, line 66: system update user is a backend user so do we need to grant history ?

mapingo: It is needed when doing the Case Details filtering of the events, see HasAosDraftedEventPredicate.  We need to know whether there is a draft-aos event and if it is before a caseworker-reissue-application event, where aosIsDrafted field will get reset.

PR #2370, opened 2022-09-15
NFDIV-1748

tomxelliott:
### Change description ###
Generate Final Order Cover Letter and configure Offline Bulk Print Notifications

https://tools.hmcts.net/jira/browse/NFDIV-1748


src/main/java/uk/gov/hmcts/divorce/caseworker/service/notification/FinalOrderGrantedNotification.java
```diff
@@ -39,6 +42,27 @@ public class FinalOrderGrantedNotification implements ApplicantNotification {
...
+            caseData.getApplicant1().getSolicitor().getEmail(),
+            SOLICITOR_FINAL_ORDER_GRANTED,
+            solicitorTemplateContent(caseData, caseId, caseData.getApplicant1(), caseData.getApplicant2()),
+            caseData.getApplicant1().getLanguagePreference()
+        );
+    }
+
+    @Override
+    public void sendToApplicant1Offline(final CaseData caseData, final Long caseId) {
+        log.info("Sending Final Order Granted letter to Applicant 1: {}", caseId);
```
amoriartyMOJ, line 62: Could we move the generation of docs from the aboutToSubmit into these methods? As there would be no need for any if checks and would clean it up (if I understand correctly 😂 )

amoriartyMOJ (approved):

src/main/java/uk/gov/hmcts/divorce/caseworker/service/notification/FinalOrderGrantedNotification.java
tomxelliott: Unfortunately any documents generated in the submitted callback are not saved, so we need to do this before the submitted.

nitinprabhu (approved):

PR #2542, opened 2022-10-19
Set conditional order default fields so that event is visible in dropdown list for solicitors

nitinprabhu:
### Change description ###

- This is a forward fix for cases where solicitors are not able to Draft CO.
- Draft CO is only visible when coIsDrafted is set to No 

https://tools.hmcts.net/jira/browse/NFDIV-3085

- [x] README and other documentation has been updated / added (if needed)


src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerNoticeOfChange.java
```diff
@@ -155,11 +156,26 @@ private CaseData correctRepresentationDetails(final CaseData data,
             data.getApplicant1().setAddress(beforeData.getApplicant1().getAddress());
             if (YES.equals(data.getNoticeOfChange().getAreTheyRepresented())) {
                 data.getApplicant2().setAddress(beforeData.getApplicant2().getAddress());
+                data.getConditionalOrder().getConditionalOrderApplicant1Questions().setIsSubmitted(NO);
```
amoriartyMOJ, line 159: this chunk of code looks identical to the private method created below (setConditionalOrderDefaultValues). Could you just call that instead?

PR #2544, opened 2022-10-19
NFDiv-2966 incorrect hub when co refused for admin clarification

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-2966


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Application.java
```diff
@@ -496,6 +496,12 @@ public class Application {
     )
     private YesOrNo switchedToSoleCo;
 
+    @CCD(
+        label = "Is a required admin clarification now submitted?",
+        access = {CaseworkerAccess.class}
+    )
```
nitinprabhu, line 502: Should this field be in ConditionalOrder class ?

BWeeks-JusticeGovUK: Relocated.

nitinprabhu (approved):

src/main/java/uk/gov/hmcts/divorce/legaladvisor/event/LegalAdvisorMakeDecision.java
```diff
@@ -136,6 +137,8 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(final CaseDet
         final CaseData caseData = details.getData();
         final ConditionalOrder conditionalOrder = caseData.getConditionalOrder();
 
+        caseData.getConditionalOrder().setIsAdminClarificationSubmitted(YesOrNo.NO);
```
benouaer, line 140: should this be set to null rather than No, since we're just resetting (I assume)

PR #2552, opened 2022-10-21
NFDIV-2537

tomxelliott:
### Change description ###
Handle scanned docs with preset Document sub types passed by Exela using QR Code

https://tools.hmcts.net/jira/browse/NFDIV-2537

https://user-images.githubusercontent.com/20407957/197512378-b5916d11-1ab1-4250-86b6-a9a2d0539f76.mov


src/main/java/uk/gov/hmcts/divorce/systemupdate/event/SystemAttachScannedDocuments.java
```diff
@@ -42,18 +68,78 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
...
+
+        if (mostRecentScannedSubtypeReceived.isPresent()) {
+            final ScannedDocument scannedDocument = mostRecentScannedSubtypeReceived.get();
+            final CaseDocuments.ScannedDocumentSubtypes scannedDocumentSubtype =
+                CaseDocuments.ScannedDocumentSubtypes.valueOf(scannedDocument.getSubtype().toUpperCase(Locale.ROOT));
+            final DocumentType documentType = getDocumentType(scannedDocumentSubtype);
+
+            if (isEmpty(documentType)) {
+                return AboutToStartOrSubmitResponse.<CaseData, State>builder()
+                    .errors(singletonList("Scanned Document Subtype is an invalid form type"))
```
nitinprabhu, line 118: I think we can't throw error while bulk scan attaches document to a case. What probably we should do is either classify it as Other or just attach it as it happens now ?

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerOfflineDocumentVerified.java
```diff
@@ -178,24 +183,35 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(CaseDetails<C
         log.info("{} about to submit callback invoked for Case Id: {}", CASEWORKER_OFFLINE_DOCUMENT_VERIFIED, details.getId());
         var caseData = details.getData();
 
-        if (AOS_D10.equals(caseData.getDocuments().getTypeOfDocumentAttached())) {
+        if (AOS_D10.equals(caseData.getDocuments().getTypeOfDocumentAttached())
+            || D10.equals(caseData.getDocuments().getScannedSubtypeReceived())) {
 
-            reclassifyScannedDocumentToChosenDocumentType(caseData, RESPONDENT_ANSWERS);
+            if (isEmpty(caseData.getDocuments().getScannedSubtypeReceived())) {
```
nitinprabhu, line 189: can we include this in above OR condition

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerOfflineDocumentVerified.java
```diff
@@ -178,24 +183,35 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(CaseDetails<C
...
+        } else if (CO_D84.equals(caseData.getDocuments().getTypeOfDocumentAttached())
+            || D84.equals(caseData.getDocuments().getScannedSubtypeReceived())) {
+
+            if (isEmpty(caseData.getDocuments().getScannedSubtypeReceived())) {
+                reclassifyScannedDocumentToChosenDocumentType(caseData, CONDITIONAL_ORDER_APPLICATION);
+            }
 
-            reclassifyScannedDocumentToChosenDocumentType(caseData, CONDITIONAL_ORDER_APPLICATION);
+            if (!SWITCH_TO_SOLE.equals(caseData.getConditionalOrder().getD84ApplicationType())) {
+                caseData.getDocuments().setScannedSubtypeReceived(null);
```
nitinprabhu, line 213: if we set this to null then we won't be able to search it from search filter ?

tomxelliott: This is because we use the value to filter D36, D10 and D84 that need to be picked up by Caseworkers to action, once actioned we don't want these cases to be filtered anymore

nitinprabhu (approved):

PR #2598, opened 2022-11-04
NFDIV-2204

tomxelliott:
### Change description ###
Point at new document api in functional tests

https://tools.hmcts.net/jira/browse/NFDIV-2204


src/main/resources/application.yaml
```diff
@@ -10,7 +10,9 @@ management:
       base-path: /
       exposure:
         include: health, info, prometheus
-
+  health:
+    case-document-am-api:
+      enabled: ${ENABLE_CDAM:true}
```
nitinprabhu, line 15: default should be false ?

tomxelliott: I have set false in aat and prod using flux

nitinprabhu (approved):

PR #2646, opened 2022-11-17
Fix failing to update bulkcase after pronouncing cases

mapingo:
### Change description ###

Fix bulkcase failing to update after pronouncing cases

https://tools.hmcts.net/jira/browse/NFDIV-3164

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/bulkaction/service/CasePronouncementService.java
+++ b/src/main/java/uk/gov/hmcts/divorce/bulkaction/service/CasePronouncementService.java
@@ -60,8 +60,8 @@ public class CasePronouncementService {
         final List<ListValue<BulkListCaseDetails>> erroredCaseDetails = bulkActionCaseData.getErroredCaseDetails();
         final List<ListValue<BulkListCaseDetails>> processedCaseDetails = bulkActionCaseData.getProcessedCaseDetails();
 
-        erroredCaseDetails.removeAll(erroredCaseDetails.stream().toList());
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/divorce/bulkaction/service/CasePronouncementService.java
```diff
@@ -57,11 +57,15 @@ public void pronounceCases(final CaseDetails<BulkActionCaseData, BulkActionState
         final User user = idamService.retrieveSystemUpdateUserDetails();
         final String serviceAuth = authTokenGenerator.generate();
 
-        filterCasesNotInCorrectState(bulkActionCaseData, user, serviceAuth);
+        final List<ListValue<BulkListCaseDetails>> erroredCaseDetails = bulkActionCaseData.getErroredCaseDetails();
+        final List<ListValue<BulkListCaseDetails>> processedCaseDetails = bulkActionCaseData.getProcessedCaseDetails();
+
+        erroredCaseDetails.removeAll(erroredCaseDetails.stream().toList());
```
FaisalMoJ, line 63: Is this meant to be `bulkActionCaseData.removeAll(...)`?

mapingo: Yes these need to be reset as the filter method (filterCasesNotInCorrectState) will add to the lists, and was causing issues with multiple occurances of the same case being processed.   These changes are the quick fix to allow the cases to be pronounced.  I will follow up with a refactor so it is not so complicated.  I should have used the method clear() really here., so will change that.

PR #2729, opened 2022-12-11
NFDIV-3223: Fix issue with service type setting

FaisalMoJ:
### Change description ###

Service type was set to Personal Service if applicant1 did not know the address of applicant2. This needs to be changed as there is no way for applicant1 to serve applicatn2 if they don't know the address.

https://tools.hmcts.net/jira/browse/NFDIV-3223

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/caseworker/service/task/SendAosPackToRespondent.java
+++ b/src/main/java/uk/gov/hmcts/divorce/caseworker/service/task/SendAosPackToRespondent.java
@@ -8,6 +8,7 @@ import uk.gov.hmcts.divorce.caseworker.service.print.AosPackPrinter;
 import uk.gov.hmcts.divorce.divorcecase.model.CaseData;
 import uk.gov.hmcts.divorce.divorcecase.model.State;
 import uk.gov.hmcts.divorce.divorcecase.task.CaseTask;
+import static uk.gov.hmcts.ccd.sdk.type.YesOrNo.NO;
 
 @Component
 @Slf4j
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/divorce/caseworker/service/task/SendAosPackToRespondent.java
```diff
@@ -22,7 +24,8 @@ public CaseDetails<CaseData, State> apply(final CaseDetails<CaseData, State> cas
         final Long caseId = caseDetails.getId();
         final CaseData caseData = caseDetails.getData();
 
-        if (caseData.getApplication().isCourtServiceMethod()) {
+        if (caseData.getApplication().isCourtServiceMethod()
+                && NO.equals(caseData.getApplication().getApplicant1WantsToHavePapersServedAnotherWay())) {
```
sunilnatikar, line 28: May be we can use 

`!YES.equals(caseData.getApplication().getApplicant1WantsToHavePapersServedAnotherWay())`

So that along with NO, this condition will be true for NULL as well

sunilnatikar (approved):

src/main/java/uk/gov/hmcts/divorce/caseworker/service/task/SendAosPackToRespondent.java
FaisalMoJ: I actually had it as you have suggested to counter the null values and discussed with Nitin the same this morning but not sure why it has reverted to `NO.equals...` 😆 i will update this now

FaisalMoJ: Changed from `NO.equals..` to `!YES.equals...` for null check as well

PR #2737, opened 2022-12-13
NFDIV-2950: added welsh template changes for JS COE solicitor docs

sunilnatikar:
### Change description ###
Added welsh template changes for JS COE solicitor docs

### JIRA link ###
https://tools.hmcts.net/jira/browse/NFDIV-2950

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/document/content/DocmosisCommonContent.java
+++ b/src/main/java/uk/gov/hmcts/divorce/document/content/DocmosisCommonContent.java
@@ -51,7 +51,7 @@ public class DocmosisCommonContent {
     @Value("${court.locations.serviceCentre.poBox}")
     private String poBox;
 
-    @Value("${court.locations.serviceCentre.poBoxCy:Blwch Post 13226}")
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/hmcts/divorce/document/content/DocmosisCommonContent.java
```diff
@@ -48,6 +51,9 @@ public class DocmosisCommonContent {
     @Value("${court.locations.serviceCentre.poBox}")
     private String poBox;
 
+    @Value("${court.locations.serviceCentre.poBoxCy:Blwch Post 13226}")
```
nitinprabhu, line 54: does it need a default ? just thinking from consistency point as `@Value("${court.locations.serviceCentre.poBox}")
     private String poBox;` doesn't has default

sunilnatikar: we can add this in app config if we don't want to mention it here as a default?

nitinprabhu (approved):

PR #2766, opened 2022-12-21
Generate apply for FO letters for applicant 2 only when they are not represented and without email

nitinprabhu:

https://tools.hmcts.net/jira/browse/NFDIV-3243


sunilnatikar (approved):

amoriartyMOJ:
Ticket was withdrawn. N/A.

PR #2834, opened 2023-02-13
NFDiv-3253 Extend "enum" field to include Judicial Separation

BWeeks-JusticeGovUK:
**** DO NOT MERGE UNTIL ALL OTHER JS TICKETS ARE MERGED ****
**** All calls to getIsJudicialSeparation() should be refactored to isJudicialSeparationCase() ****
**** All tests setting caseData.isJudicialSeparation() should be refactored to caseData.setSupplementaryCaseType() ****
**** All test JSON data including isJudicialSeparation field should be refactored to supplementaryCaseType field ****

https://tools.hmcts.net/jira/browse/NFDIV-3253

Add supplementaryCaseType enum with: NA, JUDICIAL_SEPARATION, SEPARATION, NULLITY
isJudicialSeparation() = is JUDICIAL_SEPARATION || SEPARATION (ok to check for either here - journey is identical)
isNullity() = is NULLITY
isNotApplicable() = is NA

CaseData:
Add supplementaryCaseType field with a default value of notApplicable.  This ensures all new cases gain the field, and all existing cases will gain the field once an event is triggered for that case.
Override default divorceOrDissolution setter to ensure supplementaryCaseType is set correctly for Div/JS, Diss/Sep
Override default supplementaryCaseType setter to ensure the correct JS/S type is always set based on divorceOrDissolution.
Remove isJudicialSeparation field, and refactor existing code appropriately.

Refactor isJudicialSeparationCase() to check supplementaryCaseType for JUDICIAL_SEPARATION || SEPARATION values.
Add isNullityCase() - checks supplementaryCaseType for NULLITY.
Add hasNoAdditionalCaseType() - checks supplementaryCaseType for NA or null. 

D8SFormToCaseTransformer
Refactor setIsJudicialSeparation(YES) to setSupplementaryCaseType(JUDICIAL_SEPARATION)
**** Once the D8S form is altered, refactor this to set based on the scanned marriage/civil partnership value ****

ApplicationTab
Only show divorceOrDissolution if supplementaryCaseType is NA
Only show supplementaryCaseType if supplementaryCaseType is *not* NA

PronouncementListTemplateContent
If the case has a supplementaryCaseType, populate the divorceOrDissolution field in the templateContent with supplementaryCaseType.  Otherwise, use divorceOrDissolution value.

Tests/JSON files
Update all tests and related JSON files with appropriate supplementaryCaseType values.
 
As show conditions cannot accept a null value, this PR may require a data migration to ensure that all existing cases have a supplementaryCaseType field with a value of NA.


src/test/java/uk/gov/hmcts/divorce/bulkscan/transformation/D8sFormToCaseTransformerTest.java
```diff
@@ -126,7 +126,7 @@ void shouldSuccessfullyTransformD8FormWithScannedDocumentsWithoutWarnings() thro
         final CaseData captorValueCaseData = transformationDetailsCaptorValue.getCaseData();
 
         assertThat(captorValueCaseData.getDivorceOrDissolution()).isEqualTo(DIVORCE);
-        assertThat(captorValueCaseData.getIsJudicialSeparation()).isEqualTo(YES);
+        assertThat(captorValueCaseData.isJudicialSeparationCase()).isTrue();
```
amoriartyMOJ, line 129: the .isTrue() isn't needed here is it?

src/main/java/uk/gov/hmcts/divorce/systemupdate/schedule/SystemJsDisputedAnswerOverdueTask.java
```diff
@@ -59,7 +60,8 @@ public void run() {
             final BoolQueryBuilder query =
                 boolQuery()
                     .must(matchQuery(STATE, AwaitingAnswer))
-                    .must(matchQuery(String.format(DATA, IS_JUDICIAL_SEPARATION), YesOrNo.YES))
+                    .mustNot(matchQuery(String.format(DATA, SUPPLEMENTARY_CASE_TYPE), NA))
```
amoriartyMOJ, line 63: Could these be a must(matchQuerry(String.format(DATA, SUPPLEMENTARY_CASE_TYPE), JUDICIAL_SEPARATION) ? Instead of checking for what it is not? similar to before the rework?

BWeeks-JusticeGovUK: We'd need to be able to match with an OR qualifier (it must match on js OR sep OR nullity) - I couldn't find any documentation or other examples when I looked before.  Do you know if this is possible?

BWeeks-JusticeGovUK: If it's not possible to match with OR, do we need to add an additional mostNot match for a null value?

src/main/java/uk/gov/hmcts/divorce/bulkscan/transformation/D8sFormToCaseTransformer.java
```diff
@@ -79,7 +79,7 @@ protected Map<String, Object> runFormSpecificTransformation(
 
             List<String> transformationWarnings = transformationDetails.getTransformationWarnings();
             caseData.setDivorceOrDissolution(DIVORCE);
-            caseData.setIsJudicialSeparation(YES);
+            caseData.setSupplementaryCaseType(JUDICIAL_SEPARATION);
```
amoriartyMOJ, line 82: Raised a question about how we can determine a case is SEPARATION and not just JUDICIAL_SEPARATION because currently we hard default cases to Divorce && judicial separation. So docmosis docs and other docs will never be able to determine if it should not render the "Judicial" keyword.

Raising a question on this PR but will get it fixed in a follow up PR if it is needed.

amoriartyMOJ (approved):
LGTM - Will need extensive testing by QA as mentioned on call.

Some notable areas to test along with general regression:

- Existing cases (online & paper) get a default "NA" value assigned for case type, and that all areas still work as expected with no changes to UI.

- New cases (both div, diss & JS) can be created, progressed and for each case type the application tab now shows it's correct type.

- Pronouncement works regardless of it's case type (div, diss & js). And the pronouncement list shows for JS cases that it is a judicial separation case, not just a divorce || dissolution case.

- Test that correct paper case for JS cases, if dissolution -> case type = Separation. If divorce -> case type = judicial separation.

likely more that needs testing, but these are the important few.

PR #2878, opened 2023-03-29
Update app insights to 3.4.10

BWeeks-JusticeGovUK:
Update app insights to 3.4.10 per https://tools.hmcts.net/jira/browse/DTSPO-13217

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,11 +1,20 @@
 {
-  "enabledManagers": ["helm-requirements","gradle-wrapper", "terraform"],
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "schedule": "after 7am and before 11am every weekday",
+  "extends": ["config:base"],
   "labels": ["dependencies"],
-  "helm-requirements":
-  {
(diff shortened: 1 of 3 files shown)


.github/renovate.json
```diff
@@ -1,11 +1,20 @@
 {
-  "enabledManagers": ["helm-requirements","gradle-wrapper", "terraform"],
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
```
timja, line 2: you'll need to delete .github/dependabot.yml if you take this change as is.
previously it was selectively enabling managers and now it has them all enabled

timja (approved):
to unblock critical issues this is fine anyway

Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG APP_INSIGHTS_AGENT_VERSION=3.2.6
+ARG APP_INSIGHTS_AGENT_VERSION=3.4.10
```
timja, line 1:
```suggestion
# renovate: datasource=github-releases depName=microsoft/ApplicationInsights-Java
ARG APP_INSIGHTS_AGENT_VERSION=3.4.10
```

PR #2955, opened 2023-05-19
NFDiv-3446 Remove 20 week date for CO application from Deemed Service letter

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-3446


src/main/java/uk/gov/hmcts/divorce/document/content/ServiceOrderTemplateContent.java
```diff
@@ -69,9 +76,16 @@ public Map<String, Object> apply(final CaseData caseData,
 
         templateContent.put(CASE_REFERENCE, caseData.formatCaseRef(ccdCaseReference));
 
-        final LanguagePreference applicantLanguagePreference = caseData.getApplicant1().getLanguagePreference();
+        final LanguagePreference applicantLanguagePreference = caseData.isJudicialSeparationCase()
```
amoriartyMOJ, line 79: This one can be removed and we can call the language preference directly (null / no - english)

amoriartyMOJ (approved):
Approved with 1 comment. We can remove the is JS check and instead just store the lang preference as the result of the call.

PR #3015, opened 2023-06-13
Fix Broken CoE Unit Tests

BWeeks-JusticeGovUK:
Multiple CoE Unit tests are broken as a result of using atMostOnce(), which allows tests to pass when verified functions are not called.

Refactored CoE content generator methods into a new helper class.  Updated/added tests.


src/test/java/uk/gov/hmcts/divorce/testutil/TestDataHelper.java
```diff
@@ -258,19 +275,17 @@ public static Applicant getInvalidApplicant() {
 
     public static Solicitor getOfflineSolicitor() {
         AddressGlobalUK addressGlobalUK = AddressGlobalUK.builder()
-            .addressLine1("line1")
-            .addressLine2("line2")
-            .postTown("city")
-            .postCode("postcode")
+            .addressLine1("sol line1")
```
amoriartyMOJ, line 278: As this was a pre-existing method in the helper was it used by any other classes as if it was the assertions will have changed for those?

BWeeks-JusticeGovUK: No.  I actually added that method, and I've re-tested both existing uses.  Changing the values has no effect on those tests :)

amoriartyMOJ (approved):

PR #3055, opened 2023-06-29
NFDIV-3463 INC5510831 - new feature: expediting cases

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-3463

Adds new caseworker event: caseworkerExpediteFinalOrder.

To trigger an expedited FO, a judge must provide an appropriate general order.  Therefore if the case does not contain any general orders, this event will immediately fail with an appropriate error.

When the event is successfully triggered, the caseworker is asked to select the correct general order document from those present on the case, and to manually enter the name of the authorising judge.  As per the caseworkerGrantFinalOrder event, they also need to select the 'Granted' tick box to proceed.  Completing the event will grant the final order immediately.

This selected general order and the judge name are then stored in a new final order object: expeditedFinalOrderAuthorisation.

The object contains two fields: expeditedFinalOrderGeneralOrder and expeditedFinalOrderJudgeName.

The new fields will be shown on the final order tab if the final order was expedited.

NB: This differs from the AC as a result of team conversations.  Requesting final confirmation from the business.


src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerExpediteFinalOrder.java
```diff
@@ -0,0 +1,192 @@
...
+        var caseData = details.getData();
+
+        if (caseData.getGeneralOrders() == null) {
+            return AboutToStartOrSubmitResponse.<CaseData, State>builder()
+                .data(caseData)
+                .errors(Collections.singletonList("No general order documents found.  Unable to continue."))
+                .build();
+        }
+
+        final var generalOrderDocuments = caseData.getGeneralOrders()
```
amoriartyMOJ, line 108: if the documents are null, do we want them accessing this event? Would it be better to add the check for null in the initial check of getGeneralOrders() == null above. As then we can assume if they get past that point documents do exist.

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerExpediteFinalOrder.java
```diff
@@ -0,0 +1,192 @@
...
+            .filter(g -> g.getValue().getGeneralOrderDocument().getDocumentFileName().equals(expeditedFinalOrderGeneralOrderDocumentName))
+            .findFirst();
+
+        if (generalOrderToExpediteFinancialOrder.isEmpty()) {
+            return AboutToStartOrSubmitResponse.<CaseData, State>builder()
+                .data(caseData)
+                .errors(
+                    Collections.singletonList(
+                        String.format(
+                            "Cannot locate selected general order document (%s).  Unable to continue.",
```
amoriartyMOJ, line 160: If we assert that the docs exist in the aboutToStart, is this check needed?

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/ExpeditedFinalOrderAuthorisation.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import uk.gov.hmcts.ccd.sdk.api.CCD;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+@Builder
+public class ExpeditedFinalOrderAuthorisation {
```
amoriartyMOJ, line 13: Bit of a memory block. Are these fields okay with existing cases?

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerExpediteFinalOrder.java
BWeeks-JusticeGovUK: Not sure I'm following this - we are not concerned with the documents field at all for this event.  Can we discuss please?

amoriartyMOJ (approved):
Looks good. Linus was taking a look at the big CCD change so will see if he has an update for it on Tuesday when i'm back in.

amoriartyMOJ:
Wanted to leave a comment to address the mass changes we're seeing in the CCD Report. Linus spoke to Barry and I on a call and because Barry has added a new Complex type; it's being calculated first as it's the most nested complex type found. That means that when it's generating the rest of the definition files they've all be "bumped" down the list so their names have changed from e.g. 2_Payments -> 3_Payments.

Although it doesn't fully make sense, it's expected behaviour and isn't anything to worry about for this PR.

PR #3062, opened 2023-06-30
Fix generate-ccd-definition.sh

BWeeks-JusticeGovUK:
Remove user param from docker run call in generate-ccd-definition.sh.

This prevents the inaccessible ./cache folder issue when running the definition processor.


FaisalMoJ (approved):

linusnorton (approved):
I believe I originally added this to fix an issue on Linux. I can't recreate the issue so it seems to be all good

PR #3106, opened 2023-07-19
NFDIV-3567 - Bulk Print duplicate request tracking

amoriartyMOJ:
### Change description ###

A new key added to the additional_information 'recipients' which has a value of List<String>. This list is a unique identifier for each request that is made to bulk print. It will be unique per request and repeatable should any request retry.

The List<String> recipients should always = List.of(case reference number, recipients full name, letter pack type).
Example in AosOverduePrinter:  List.of(caseIdString, recipient.getFullName(), LETTER_TYPE_AOS_OVERDUE)

Bulk print will monitor for duplicate requests within an hour period and if any are found it will reject them.

### IMPORTANT DEV NOTE ###
The Print class temporarily has been given 2 constructors. The first will be an all arguments constructor which is what we will use in the future, passing in the List<String> recipients.

The other constructor doesn't take a List<String> recipients, and instead sets it automatically to a random string. This is to avoid compilation errors and failing functional tests (409 conflicts) until all Printer classes have been migrated to the new constructor. Please do not merge this PR with this constructor still existing!

What's been done?
- The core change to the bulk print service & Print model have been complete.
- An example migration has been made for the AoS overdue notification to use the new field.

Work still outstanding:

- Rework all places which call to print documents to include a recipient.
- Rework all unit tests to account for changes.

https://tools.hmcts.net/jira/browse/NFDIV-3567


src/main/java/uk/gov/hmcts/divorce/document/print/model/Print.java
```diff
@@ -12,4 +12,6 @@ public class Print {
     private final String caseId;
     private final String caseRef;
     private final String letterType;
+    private final List<String> recipients;
```
nigeldunne, line 15: Consider removing the @AllArgsConstructor and creating the list in the constructor

    public Print(final List<Letter> letters, final String caseId, final String caseRef, final String letterType, final String recipientName) {
        this.letters = letters;
        this.caseId = caseId;
        this.caseRef = caseRef;
        this.letterType = letterType;
        this.recipients = List.of(caseId, recipientName, letterType);
    }

PR #3221, opened 2023-09-19
Use CCD in AAT for PRs

linusnorton:
### Change description ###

This change removes the CCD stack from preview in favour of using it in AAT.

A new case type is created with the PR number in so that it doesn't conflict with the case type in AAT, or any other PRs being built.

The callbacks are configured to use nfdiv-case-api in Preview. 

You should use XUI in AAT to do any manual testing. The nfdiv-frontend will be built in Preview as normal. 

Manage Orgs will not work as it contains a hardcoded list of case types.


Jenkinsfile_CNP
```diff
@@ -228,6 +238,17 @@ withPipeline(type, product, component) {
...
+  afterAlways('functionalTest:preview') {
+    if (!checkForKeepHelmLabel(env.BRANCH_NAME)) {
+      sh "./bin/remove-case-type.sh"
+    }
+  }
+
+}
+
+def checkForKeepHelmLabel(branch_name) {
+  return new GithubAPI(this).getLabelsbyPattern(branch_name, "enable_helm_keep").contains("enable_helm_keep")
```
timja, line 251:
```suggestion
  return new GithubAPI(this).checkForLabel(env.BRANCH_NAME, "enable_keep_helm")
```

src/test/java/uk/gov/hmcts/divorce/divorcecase/NoFaultDivorceTest.java
```diff
@@ -19,12 +19,12 @@ public class NoFaultDivorceTest {
...
 
     @Test
-    void shouldAddSystemUpdateUserAccessToDraftStateWhenEnvironmentIsAat() {
+    void shouldSetTheCorrectCaseTypeName() {
         final ConfigBuilderImpl<CaseData, State, UserRole> configBuilder = createCaseDataConfigBuilder();
 
         noFaultDivorce.configure(configBuilder);
 
-        assertThat(configBuilder.build().getCaseType()).isEqualTo("NFD");
+        assertThat(configBuilder.build().getCaseDesc()).isEqualTo("Handling of the dissolution of marriage");
```
velimd, line 27: test says correct case type but this is comparing case description. is that correct?

linusnorton: this is a bit of a fudge as I wasn't able to construct a property test as System.getenv() is immutable

velimd (approved):

PR #3318, opened 2023-11-02
NFD - Super user event to hard delete docs from case data and case event history

amoriartyMOJ:
### Change description ###

This pull request aims to rework the super user event for document removal, to do a hard deletion instead of a soft delete. This should allow us to delete the document from case history as well as current case data.

Our aim is to cut down the amount of incidents where documents need to be removed from the case history tab after incorrect uploading from caseworkers and users.

https://tools.hmcts.net/jira/browse/NFDIV-TBC


src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerRemoveDocument.java
```diff
@@ -36,4 +50,51 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
...
+            beforeDocumentsGenerated.forEach(document -> {
+                if (!currentDocumentsGenerated.contains(document)) {
+                    documentsToRemove.add(document);
+                }
+            });
+        }
+
+        var beforeDocumentsUploaded = beforeCaseData.getDocuments().getDocumentsUploaded();
+        var currentDocumentsUploaded = currentCaseData.getDocuments().getDocumentsUploaded();
+        if (beforeDocumentsUploaded != null & currentDocumentsUploaded != null) {
```
linusnorton, line 84: that's three pretty similar blocks of code, perhaps there could be a method that will diff two lists of documents?

src/main/java/uk/gov/hmcts/divorce/document/DocumentRemovalService.java
```diff
@@ -0,0 +1,46 @@
...
+    @Autowired
+    private IdamService idamService;
+
+    public void deleteDocumentFromDocumentStore(final List<ListValue<DivorceDocument>> documentsToRemove) {
+
+        final User systemUser = idamService.retrieveSystemUpdateUserDetails();
+        final UserDetails userDetails = systemUser.getUserDetails();
+        final String rolesCsv = String.join(",", userDetails.getRoles());
+
+        documentsToRemove.forEach(document -> {
```
linusnorton, line 35: not sure what the performance is like but if there is a large number of docs it might be worth doing .stream().parallel()

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerRemoveDocument.java
amoriartyMOJ: Implemented in latest push

src/main/java/uk/gov/hmcts/divorce/document/DocumentRemovalService.java
amoriartyMOJ: implemented, thanks!

linusnorton (approved):

velimd (approved):

PR #3348, opened 2023-11-14
Remove dm-store and use cdam

linusnorton:
### Change description ###

Switch over from dm-store to CDAM. 

Fix the CaseworkerIssueApplication and CaseworkerReissueApplication integration tests so that they correctly check for when a D84 and D10 have been attached to the case.

Disable all the functional tests that involve an event that would upload a document as CDAM expects a CCD case to exist and our functional tests use a dummy case ID


charts/nfdiv-case-api/values.yaml
```diff
@@ -45,7 +45,7 @@ java:
     CASE_DATA_STORE_BASEURL: "http://ccd-data-store-api-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
     AAC_MANAGE_CASE_ASSIGNMENT_BASEURL: "http://aac-manage-case-assignment-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal"
     DOC_ASSEMBLY_URL: http://dg-docassembly-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal
-    DOCUMENT_MANAGEMENT_URL: http://dm-store-{{ .Values.global.environment }}.service.core-compute-{{ .Values.global.environment }}.internal
```
nigeldunne, line 48: Does this need to be updated in the flux for the cron jobs?

src/main/java/uk/gov/hmcts/divorce/document/DocumentManagementClient.java
```diff
@@ -18,7 +18,7 @@
 import static uk.gov.hmcts.divorce.document.DocumentConstants.USER_ID;
 import static uk.gov.hmcts.divorce.document.DocumentConstants.USER_ROLES;
 
-@FeignClient(name = "document-management-api", url = "${document_management.url}")
+@FeignClient(name = "document-management-api", url = "${case_document_am.url}")
```
nigeldunne, line 21: Should we update the client name?

charts/nfdiv-case-api/values.yaml
linusnorton: It will need to go in for both the nfdiv-case-api and the crons

src/main/java/uk/gov/hmcts/divorce/document/DocumentManagementClient.java
nigeldunne: I mean this

linusnorton: The whole file has been deleted now

nigeldunne (approved):

PR #3384, opened 2023-11-21
Re-enable previously failing FT to test in preview

amoriartyMOJ:
### Change description ###

It was failing due to a move to CDAM causing issues with us using fake ccd cases and document generation. Testing if this is still an issue


amoriartyMOJ:
Although some tests were fixed, it seems many are still broken. This is due to cdam needing a real ccd case id for documents. I've tried creating another branch which uses a real ccd case id but the pre-made request we fire into the FTs which contain existing documents, fail as they don't technically belong to the case.

Closing this PR. Going forward we will likely move to playwright and make these tests priority for first written tests.

PR #3388, opened 2023-11-21
Option 2: Fix/unset fo sot when overdue 

velimd:
### Change description ###
Nightly test fail because the the fo sot of field is set to true before the case goes overdue. This makes sure the fields are set to null when overdue.

Option 1: https://github.com/hmcts/nfdiv-frontend/pull/3368

https://tools.hmcts.net/jira/browse/NFDIV-

**Before raising**

**Before merging**


velimd:
Closing because I dont think this makes sense. The issue is we dont know if the other app will now apply even if its overdue so we cant set both to null.

PR #3400, opened 2023-11-24
NFDIV-3824 - Lock down cases in bulk list flow

amoriartyMOJ:
### Change description ###

Create a new state specifically for cases which are in bulk lists.

Use the new state to make sure cases that are linked to a bulk list, can't be easily changed unless they are first removed from the bulk list.

Update the caseworker-drop-case-from-bulk-list event to move a case back to AwaitingPronouncement if a caseworker drops it from a bulk list.This is to allow the cron which picks cases up and adds them to bulk lists, to find the case if no other actions are taken against it. If a case has been pronounced or progressed further, don't move the state back when it's dropped.

https://tools.hmcts.net/jira/browse/NFDIV-3824


amoriartyMOJ:
Needs more thought on how AttachScannedDocuments & OfflineDocumentVerified events would be handled? As if we deny access while in bulk list, would this cause issues? 

A possibility could be to change the offline document verified event to first check if a case is in a bulk list. If it is, we remove it before the event can move to other states?

amoriartyMOJ:
To update on my previous comment:

> Needs more thought on how AttachScannedDocuments & OfflineDocumentVerified events would be handled? As if we deny access while in bulk list, would this cause issues?
> 
> A possibility could be to change the offline document verified event to first check if a case is in a bulk list. If it is, we remove it before the event can move to other states?

To update on this previous comment ^
We believe we can allow the AttachScannedDocuments & OfflineDocumentsVerified flow to be accessible to caseworker. It's unlikely that they would action an A10 (AOS response) after it's already made it to the bulk flow, due to the fact you need an AOS before you get there. Same goes for the Final order & other types. 

This PR will allow the flow to be accessible as it was previously, but aim to lock down other events.

src/main/java/uk/gov/hmcts/divorce/bulkaction/task/PronounceCasesTask.java
```diff
@@ -15,13 +15,14 @@
...
 import static uk.gov.hmcts.divorce.divorcecase.model.State.ConditionalOrderPronounced;
+import static uk.gov.hmcts.divorce.divorcecase.model.State.InBulkList;
 import static uk.gov.hmcts.divorce.divorcecase.model.State.OfflineDocumentReceived;
 
 @Slf4j
 @Component
 public class PronounceCasesTask implements BulkCaseTask {
 
-    final EnumSet<State> awaitingPronouncement = EnumSet.of(AwaitingPronouncement, OfflineDocumentReceived);
+    final EnumSet<State> awaitingPronouncement = EnumSet.of(AwaitingPronouncement, OfflineDocumentReceived, InBulkList);
```
linusnorton, line 25: Do we want a TODO: to remove AwaitingPronouncement when all the existing cases are out of that state?

src/main/java/uk/gov/hmcts/divorce/bulkaction/task/RetryPronounceCasesTask.java
```diff
@@ -15,13 +15,20 @@
...
+import static uk.gov.hmcts.divorce.divorcecase.model.State.InBulkList;
 import static uk.gov.hmcts.divorce.divorcecase.model.State.OfflineDocumentReceived;
 
 @Slf4j
 @Component
 public class RetryPronounceCasesTask implements BulkCaseTask {
 
-    final EnumSet<State> awaitingPronouncement = EnumSet.of(AwaitingPronouncement, OfflineDocumentReceived, ConditionalOrderPronounced);
+    final EnumSet<State> awaitingPronouncement = EnumSet.of(
+        AwaitingPronouncement,
```
linusnorton, line 26: ditto

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/State.java
```diff
@@ -259,6 +259,13 @@ public enum State {
     )
     AwaitingJsNullity,
 
+    @CCD(
+        label = "Case in bulk list process",
+        hint = "### Case number: ${hyphenatedCaseRef}\n ### ${applicant1LastName} and ${applicant2LastName}\n",
+        access = {DefaultStateAccess.class}
+    )
+    InBulkList,
```
linusnorton, line 267: InBulkActionCase? Bit verbose but technically correct.

src/main/java/uk/gov/hmcts/divorce/systemupdate/event/SystemUpdateCaseWithCourtHearing.java
```diff
@@ -45,7 +46,7 @@ public class SystemUpdateCaseWithCourtHearing implements CCDConfig<CaseData, Sta
     public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuilder) {
         new PageBuilder(configBuilder
             .event(SYSTEM_UPDATE_CASE_COURT_HEARING)
-            .forStates(AwaitingPronouncement, OfflineDocumentReceived)
+            .forStates(AwaitingPronouncement, InBulkList, OfflineDocumentReceived)
```
linusnorton, line 49: Need a todo to remove AwaitingPronouncement

src/main/java/uk/gov/hmcts/divorce/systemupdate/event/SystemUpdateCaseWithPronouncementJudge.java
```diff
@@ -26,7 +27,7 @@ public class SystemUpdateCaseWithPronouncementJudge implements CCDConfig<CaseDat
     public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuilder) {
         new PageBuilder(configBuilder
             .event(SYSTEM_UPDATE_CASE_PRONOUNCEMENT_JUDGE)
-            .forStates(AwaitingPronouncement, OfflineDocumentReceived)
+            .forStates(AwaitingPronouncement, InBulkList, OfflineDocumentReceived)
```
linusnorton, line 30: ditto

amoriartyMOJ:
@linusnorton I've addressed your comments.
- Updated the state name to be more descriptive.
- Added a TODO comment in all places where a clean-up will be needed in the future to remove AwaitingPronouncement state.

linusnorton (approved):

PR #3485, opened 2023-12-20
Allow moving out of AwaitingApplicant1Response state for stuck cases

amoriartyMOJ:
### Change description ###

Cases stuck in awaiting app1 response can now be moved out using this event.


amoriartyMOJ:
Closed as we opted for SQL query to fix the 1 broken case. Better to not open this event up further.

PR #3528, opened 2024-01-25
Revert "NFDIV-3824 - Lock down cases in bulk list flow"

amoriartyMOJ:
As a precaution i'm creating a revert PR ready to go if anything fails with this and it needed to be pulled out.

Reverts hmcts/nfdiv-case-api#3400


amoriartyMOJ:
PR went well. Closing as not needed.

PR #3531, opened 2024-01-25
Don't send AOS notifications if case is in Awaiting Conditional Order

amoriartyMOJ:
### Change description ###

As case is already in Awaiting CO, we don't need to notify them of next steps.

Prod error.


amoriartyMOJ:
MIght need more thought, as the submit AOS service can alter the state of an application, moving it back to Holding. Faisal is looking into that though, so this PR will remain open incase it's needed.

amoriartyMOJ:
Closing as it seems this isn't needed. If an issue ever arises again, they should be able to find this PR with the closed PR search.

PR #3552, opened 2024-02-06
NFDIV-3878 Overseas Citizen Address - Backend

BWeeks-JusticeGovUK:
NFDIV-3878 https://tools.hmcts.net/jira/browse/NFDIV-3878


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Applicant.java
```diff
@@ -263,44 +259,49 @@ public String getCorrespondenceAddress() {
...
                 .collect(joining("\n"));
+        } else {
+            return Stream.of(
+                    address.getAddressLine1(),
+                    address.getAddressLine2(),
+                    address.getAddressLine3(),
+                    address.getPostTown(),
+                    address.getCounty(),
+                    address.getCountry(),
+                    address.getPostCode()
```
DelythJustice, line 270: postcode is after country but the test is checking for UK then postcode ?

PR #3562, opened 2024-02-08
NFDIV-3878 Overseas Citizen Address - Backend

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-3878


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Applicant.java
```diff
@@ -263,44 +259,49 @@ public String getCorrespondenceAddress() {
...
         if (isRepresented()) {
-            return solicitor.getAddress();
+            return Stream.of(
+                    Optional.ofNullable(solicitor.getOrganisationPolicy())
+                        .map(OrganisationPolicy::getOrganisation).map(Organisation::getOrganisationName).orElse(null),
+                    solicitor.getAddress()
+                ).filter(value -> value != null && !value.isEmpty())
+                .collect(joining("\n"));
+        } else if (null != address) {
+            return getApplicantAddress();
```
DelythJustice, line 302: changing from using AddressUtil.getPostalAddress to getApplicantAddress means we don't have anything to cope with making the address fit in the envelope window moving forward

BWeeks-JusticeGovUK: We're barely using that method anyway now - the letterPrinter handles everything moved to a docPack, and it doesn't touch it.  Whenever we wrap a long line we push the postcode down.  This was partially compensated for by leaving out the country line for UK addresses, but we won't be doing that moving forward.  It runs the risk of pushing the postcode off the bottom of the window, and the postcode is the most important part of the address.

adamg-hmcts (approved):

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Applicant.java
```diff
@@ -263,44 +259,54 @@ public String getCorrespondenceAddress() {
...
+                    address.getPostCode()
+                )
+                .filter(value -> value != null && !value.isEmpty())
+                .collect(joining("\n"));
         }
+    }
 
-        return null;
+    @JsonIgnore
+    public YesOrNo getCorrespondenceAddressOverseas() {
```
DelythJustice, line 278: could this have an Is in it so it's obvious it is a flag? Like CorrespondenceAddressIsOverseas

src/main/java/uk/gov/hmcts/divorce/document/print/model/Print.java
```diff
@@ -11,19 +12,22 @@ public class Print {
     private final String caseRef;
     private final String letterType;
     private final List<String> recipients;
+    private final Boolean isInternational;
```
DelythJustice, line 15: I think from the code setting it that we are expecting this to be boolean, ie not null just either true or false

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Applicant.java
BWeeks-JusticeGovUK: Yep.  Done.

src/main/java/uk/gov/hmcts/divorce/document/print/model/Print.java
BWeeks-JusticeGovUK: Correct.  The constructor should be passed a YesOrNo, which it will convert into a Boolean.  However it is possible for the constructor to receive a null value in place of the YesOrNo, in which case it will treat that value as if it is False.

DelythJustice: I just meant it could be boolean not Boolean is all

DelythJustice (dismissed):

amoriartyMOJ (approved):

DelythJustice (approved):

adamg-hmcts (approved):

PR #3597, opened 2024-03-01
NFDIV-3879 Overseas Solicitor Address 

BWeeks-JusticeGovUK:
NFDIV-3879
https://tools.hmcts.net/jira/browse/NFDIV-3879


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Solicitor.java
```diff
@@ -94,4 +104,52 @@ public boolean hasOrgId() {
...
+                    .replace("\n\n", "\n")
+                    .replace(",,", ",")
+                    .replace("\n,\n", "\n")
+                    .replace(", , ", ", ")
+                    .replace(" \n, \n", " \n")
+                    .replace("\n ,\n", "\n")
+                    .replace(", \n ,", ",")
+                    .replace(" \n  \n", " \n")
+                    .replace("\n \n ", "\n ")
+                    .replace(" \n \n", " \n")
```
adamg-hmcts, line 124: Maybe we could use a couple of regular expressions to simplify this matching? Something like this (probably will need to be tweaked):
```suggestion
                    .replaceAll("\n[, ]*\n", "\n")
                    .replaceAll(",[\n ]*,", ",")
```

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Solicitor.java
```diff
@@ -94,4 +104,52 @@ public boolean hasOrgId() {
         }
         return false;
     }
+
+    public String getAddress() {
+        if ((YesOrNo.NO == getAddressOverseas() || null == getAddressOverseas()) && (this.address != null && !this.address.equals(""))) {
```
adamg-hmcts, line 109: It would be nice if we could turn this into a guard clause to avoid wrapping the entire method in a block:
```suggestion
        if (getAddressOverseas() == YesOrNo.YES || this.address == null || this.address.equals("")) {
          return this.address;
        }
```
I think we would also need to check for the match against `ukPostcodeRegex` though

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Solicitor.java
```diff
@@ -94,4 +104,52 @@ public boolean hasOrgId() {
...
+                    } else if (upToFirstNewLine.endsWith(", ")
+                        && (!addressPostcodeStripped.endsWith(",") && !addressPostcodeStripped.endsWith(", "))) {
+                        addressPostcodeStripped += ", ";
+                    } else if (upToFirstNewLine.endsWith(" ") && !addressPostcodeStripped.endsWith(" ")) {
+                        addressPostcodeStripped += " ";
+                    }
+                    addressPostcodeStripped += "\n";
+                    if (addressPostcodeStripped.substring(0, upToFirstNewLine.length() + 2).endsWith(" ")) {
+                        addressPostcodeStripped += " ";
+                    }
```
adamg-hmcts, line 140: Maybe we could use a regex for this too 🤔. We could check for a match against [, ] at the end of `upToFirstNewLine` and then append to the postcode stripped address if it isn't there already

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Solicitor.java
```diff
@@ -94,4 +104,52 @@ public boolean hasOrgId() {
...
+                if (addressPostcodeStripped.contains("\n")) {
+                    String upToFirstNewLine = addressPostcodeStripped.substring(0, addressPostcodeStripped.indexOf("\n"));
+                    if (upToFirstNewLine.endsWith(",") && !addressPostcodeStripped.endsWith(",")) {
+                        addressPostcodeStripped += ",";
+                    } else if (upToFirstNewLine.endsWith(", ")
+                        && (!addressPostcodeStripped.endsWith(",") && !addressPostcodeStripped.endsWith(", "))) {
+                        addressPostcodeStripped += ", ";
+                    } else if (upToFirstNewLine.endsWith(" ") && !addressPostcodeStripped.endsWith(" ")) {
+                        addressPostcodeStripped += " ";
+                    }
```
adamg-hmcts, line 136: Maybe we could use a regex for this too 🤔. We could check for a match against [, ] at the end of `upToFirstNewLine` and then append to the postcode stripped address if it isn't there already

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Solicitor.java
```diff
@@ -94,4 +104,52 @@ public boolean hasOrgId() {
...
+                    && !addressPostcodeStripped.endsWith(", ")) {
+                    addressPostcodeStripped += ", ";
+                }
+                addressPostcodeStripped += this.address.endsWith("\n") ? postcode + "\n" : postcode;
+                return addressPostcodeStripped;
+            }
+        }
+
+        return this.address;
+    }
```
adamg-hmcts, line 154: Question about approach: Do we have to remove the postcode from its original position or would it be okay to leave the original address as it is and append a new line at the end with the postcode? It might look weird to repeat the postcode, but then we won't need to worry about removing useful information or breaking the formatting of the address

BWeeks-JusticeGovUK: I think we're going to have to move it, but I have raised the question.

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Solicitor.java
BWeeks-JusticeGovUK: Yeah we'd still have to run the matcher and default to returning the field as provided if there's no match.  I don't know if it would provide much benefit in this instance.

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Solicitor.java
BWeeks-JusticeGovUK: I have had a chat with Nicola and Kate.  The problem with duplicating the postcode is that we add a new line to the address.  This may mean that the duplicated postcode is below the bottom of the address window.  That will mean we don't qualify for reduced postage rates (postcode must be the last visible line).  I came up with an alternative which has been approved - for UK addresses we're going to take the address as provided up until the end of the postcode, then strip the rest from the response.  This whole method is about to get a lot tidier and smaller!

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Solicitor.java
BWeeks-JusticeGovUK: No longer necessary - but worked first time in my test scenarios!  nice one :)  I think you just became our regex goto ;)

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/Solicitor.java
```diff
@@ -94,4 +104,17 @@ public boolean hasOrgId() {
         }
         return false;
     }
+
+    public String getAddress() {
+        if (YesOrNo.YES == getAddressOverseas() || this.address == null || this.address.equals("")) {
```
FaisalMoJ, line 109: `this.address == null || this.address.equals("")` can be replaced with StringUtils.isEmpty(this.address)

adamg-hmcts (approved):

amoriartyMOJ (approved):

amoriartyMOJ (approved):

DelythJustice (approved):

PR #3624, opened 2024-03-18
NFDIV-3935

DelythJustice:
### Change description ###

Enter a description.

NFDIV-3935


src/main/java/uk/gov/hmcts/divorce/systemupdate/schedule/conditionalorder/SystemRedoPronouncedCoverLettersTask.java
```diff
@@ -0,0 +1,123 @@
...
+    @Autowired
+    private AuthTokenGenerator authTokenGenerator;
+
+    @Autowired
+    private ObjectMapper objectMapper;
+
+    @Autowired
+    private CaseIdChecker caseIdChecker;
+
+    public static final String NOTIFICATION_FLAG = "coverLetterResent";
```
adamg-hmcts, line 57: Do we need to use the same flag as in [SystemResendCOPronouncedCoverLettersTask](https://github.com/hmcts/nfdiv-case-api/blob/a9e8fef88b0116e60330d6567f29920d5a81d8f9/src/main/java/uk/gov/hmcts/divorce/systemupdate/schedule/conditionalorder/SystemResendCOPronouncedCoverLettersTask.java#L60)? It gets [set after sending the regenerated cover letter here](https://github.com/hmcts/nfdiv-case-api/blob/a9e8fef88b0116e60330d6567f29920d5a81d8f9/src/main/java/uk/gov/hmcts/divorce/common/service/task/SendRegeneratedCOPronouncedCoverLetters.java#L33):
```suggestion
    public static final String NOTIFICATION_FLAG = "coPronouncedCoverLetterResent";
```

DelythJustice: yes, I've now checked that the flag isn't set for any of the affected cases too so it will be valid to check this for them

src/main/java/uk/gov/hmcts/divorce/systemupdate/service/task/RegenerateConditionalOrderPronouncedCoverLetter.java
```diff
@@ -25,54 +25,60 @@ public CaseDetails<CaseData, State> apply(CaseDetails<CaseData, State> caseDetai
 
         final Long caseId = caseDetails.getId();
         final CaseData caseData = caseDetails.getData();
-
         if (caseData.getApplicant1().isApplicantOffline() && caseData.getApplicant1().isConfidentialContactDetails()) {
+            removeAndRegenerateApplicant1(caseId, caseData, coverLetterHelper);
+        }
+        if (caseData.getApplicant1().isApplicantOffline() && caseData.getApplicant2().isConfidentialContactDetails()) {
```
adamg-hmcts, line 31: Should it be applicant 2 here?
```suggestion
        if (caseData.getApplicant2().isApplicantOffline() && caseData.getApplicant2().isConfidentialContactDetails()) {
```

src/main/java/uk/gov/hmcts/divorce/systemupdate/service/task/RegenerateConditionalOrderPronouncedCoverLetter.java
```diff
@@ -25,54 +25,60 @@ public CaseDetails<CaseData, State> apply(CaseDetails<CaseData, State> caseDetai
...
-                        caseData,
-                        caseId,
-                        caseData.getApplicant2(),
-                        CONDITIONAL_ORDER_GRANTED_COVERSHEET_APP_2
-                    );
-                }
-
-                caseData.getApplicant2().setCoPronouncedCoverLetterRegenerated(YES);
-            }
+            caseData.getApplicant2().setCoPronouncedCoverLetterRegenerated(YES);
```
adamg-hmcts, line 77: Do we need to make this flag match the notification flag used [in the query](https://github.com/hmcts/nfdiv-case-api/pull/3624/files#diff-b204569affa343ecc41a9ac3342cf50199f991bbd4eea5b1f1dd42279cb35936R32)? 🤔

DelythJustice: this class is the original, I've just needed to refactor it a bit because sonar didn't like the duplication of code. That flag is the one checked later to see if something actually needs resending I think

adamg-hmcts (approved):

adamg-hmcts (approved):

adamg-hmcts (dismissed):

src/main/java/uk/gov/hmcts/divorce/systemupdate/service/task/ConditionalOrderPronouncedCoverLetterHelper.java
```diff
@@ -185,7 +186,12 @@ private Map<String, Object> templateVarsForOfflineRespondent(final CaseData case
 
         templateContent.put(DATE_OF_HEARING, dateOfHearing);
         templateContent.put(TIME_OF_HEARING, timeOfHearing);
-        templateContent.put(COURT_NAME, caseData.getConditionalOrder().getCourt().getLabel());
+        final ConditionalOrder conditionalOrder = caseData.getConditionalOrder();
+        String courtLabel = null;
```
amoriartyMOJ, line 190: I dont believe we need to set to null first here. If the label is empty it will be null anyway?

src/main/java/uk/gov/hmcts/divorce/systemupdate/service/task/RegenerateConditionalOrderPronouncedCoverLetter.java
```diff
@@ -25,54 +25,60 @@ public CaseDetails<CaseData, State> apply(CaseDetails<CaseData, State> caseDetai
 
         final Long caseId = caseDetails.getId();
         final CaseData caseData = caseDetails.getData();
-
         if (caseData.getApplicant1().isApplicantOffline() && caseData.getApplicant1().isConfidentialContactDetails()) {
```
amoriartyMOJ, line 28: Could you explain why the isConfidentialContactDetails() check happens here? Don't we need to regenerate for all broken cases?

DelythJustice: that's just me refactoring the existing class because sonar complained - I did want to leave it alone but there was too much duplication, so I've carefully moved the code to statics in this one and then call from the other.

amoriartyMOJ (approved):
Approved with 1 minor comment about the templateContent labels

PR #3643, opened 2024-03-25
NFDIV-3935 207 cases with confidential docs need to get regenrated an…

DelythJustice:
…d resent, missed in the last run because the code just didn't regenerate if confidential

### Change description ###

Enter a description.

https://tools.hmcts.net/jira/browse/NFDIV-3935

**Before raising**

**Before merging**


FaisalMoJ (approved):

src/main/java/uk/gov/hmcts/divorce/systemupdate/schedule/conditionalorder/SystemRedoPronouncedCoverLettersTask.java
```diff
@@ -56,7 +56,7 @@ public class SystemRedoPronouncedCoverLettersTask implements Runnable {
     @Autowired
     private AuthTokenGenerator authTokenGenerator;
 
-    public static final String NOTIFICATION_FLAG = "coPronouncedCoverLetterResent";
+    public static final String NOTIFICATION_FLAG = "coPronouncedConfidentialCoverLetterResent";
```
adamg-hmcts, line 59: Out of interest, how come we need to change the notification flag? I'd thought we didn't make changes for those cases in the last run and the previous flag would still be false for them

adamg-hmcts (approved):
Already approved by Faisal, but I've given it a second approval to show that the question I left about the notification flag is not a blocker

src/main/java/uk/gov/hmcts/divorce/systemupdate/schedule/conditionalorder/SystemRedoPronouncedCoverLettersTask.java
DelythJustice: It's for when there would be one confidential applicant, then the case would have run on the normal doc but not the other one so needs to run again.

PR #3684, opened 2024-04-15
[NFDIV-3972] Only show contact details through the confidential tabs (PoC)

adamg-hmcts:
### Change description ###
**NOTE:** I think the service team would like to keep the contact information within the application tab ([see my other PR](https://github.com/hmcts/nfdiv-case-api/pull/3702/files)) rather than merge this.

EXUI show conditions do not stop case fields from being included in the response that users receive, meaning that confidential addresses (controlled by show conditions in the application tab) can be viewed in browser developer tools. To stop this, we've removed information that may be confidential from the application tab and now only show it in the confidential tabs, which have role-based access.

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/divorcecase/tab/ApplicationTab.java
+++ b/src/main/java/uk/gov/hmcts/divorce/divorcecase/tab/ApplicationTab.java
@@ -113,9 +113,6 @@ public class ApplicationTab implements CCDConfig<CaseData, State, UserRole> {
             .label("LabelApplicant1DetailsAreConfidential-Heading",
                 "applicant1ContactDetailsType=\"private\"",
                 "#### ${labelContentTheApplicantOrApplicant1UC}'s contact details are confidential")
-            .field("applicant1PhoneNumber", APPLICANT_1_CONTACT_DETAILS_PUBLIC)
-            .field("applicant1Email", APPLICANT_1_CONTACT_DETAILS_PUBLIC)
-            .field("applicant1Address", APPLICANT_1_CONTACT_DETAILS_PUBLIC)
             .field("applicant1CannotUpload")
             .field("applicant1CannotUploadSupportingDocument")
             .field("applicant1KnowsApplicant2Address",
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/hmcts/divorce/divorcecase/tab/ApplicationTab.java
```diff
@@ -170,9 +167,6 @@ private void addApplicant2(final Tab.TabBuilder<CaseData, UserRole> tabBuilder)
             .label("LabelApplicant2DetailsAreConfidential-Heading",
                 "applicant2ContactDetailsType=\"private\"",
                 "#### ${labelContentTheApplicant2UC}'s contact details are confidential")
-            .field("applicant2PhoneNumber", APPLICANT_2_CONTACT_DETAILS_PUBLIC)
-            .field("applicant2Email", APPLICANT_2_CONTACT_DETAILS_PUBLIC)
-            .field("applicant2Address", APPLICANT_2_CONTACT_DETAILS_PUBLIC)
```
adamg-hmcts, line 175: Initially I looked to add role-based access to specific fields within the application tab, but it doesn't seem to be supported by CCD config generator (the entire tab is hidden).

src/main/java/uk/gov/hmcts/divorce/divorcecase/tab/CaseTypeTab.java
```diff
@@ -231,7 +230,7 @@ private void buildConfidentialApplicantTab(ConfigBuilder<CaseData, State, UserRo
     private void buildConfidentialRespondentTab(ConfigBuilder<CaseData, State, UserRole> configBuilder) {
         configBuilder.tab("ConfidentialRespondent", "Confidential Respondent")
             .forRoles(CASE_WORKER, LEGAL_ADVISOR, JUDGE, APPLICANT_2_SOLICITOR, SUPER_USER)
-            .showCondition("applicant2ContactDetailsType=\"private\" AND applicationType=\"soleApplication\"")
+            .showCondition("applicationType=\"soleApplication\"")
```
adamg-hmcts, line 233: EXUI users who have permission to access contact details would need to start viewing it through the confidential tabs, which will be available by default, regardless of whether contact details are set to private. If we agree this approach looks reasonable, I'll ask for approval from the service team, as they may need to inform CTSC that they will need to start viewing this tab for address information

pallavijustice (approved):
approved

PR #3711, opened 2024-04-30
Revert "NFDIV-4016 introduce variable for dispute fee new template"

DelythJustice:
Reverts hmcts/nfdiv-case-api#3710
I noticed that the fee0233 had not gone up in demo env https://fees-register.demo.platform.hmcts.net/fees/FEE0233 and not gone up in prod right now either


DelythJustice:
decided not to revert having the application level variable

PR #3772, opened 2024-06-05
[NFDIV-4079] Set default org policies on case submission

adamg-hmcts:
### Change description ###

For notice of change to work, we need all cases to have a default organisation policy that includes case-assigned role and organisation fields. Changes in this PR:
- Added new case task to set default organisation policy fields on case submission. This means getSolicitor() and getSolicitor().getOrganisationPolicy() will no longer be null by default for post-submission cases.
- Adjusted conditional order code that was using `applicant.getSolicitor() != null` to determine whether an applicant was represented.
- Adjusted caseworker notice of change to use organisation ID when determining whether a sol has been added or removed, rather than checking for the presence of an organisation policy

https://tools.hmcts.net/jira/browse/NFDIV-4079


src/main/java/uk/gov/hmcts/divorce/common/service/task/SetDefaultOrganisationPolicies.java
```diff
@@ -0,0 +1,64 @@
...
+            Optional.ofNullable(organisationPolicy.getOrgPolicyCaseAssignedRole())
+                .orElse(solicitorRole)
+        );
+
+        organisationPolicy.setOrganisation(
+            Optional.ofNullable(organisationPolicy.getOrganisation())
+                .orElse(new Organisation(null, null))
+        );
+    }
+}
```
adamg-hmcts, line 64: The aim is to have a reusable task that only populates the fields which are missing. Hopefully we can then run it safely for any case and reuse the task for backfilling or later in the case sequence (e.g. after caseworker NoC).

DelythJustice (approved):

DelythJustice (approved):

PR #3837, opened 2024-07-03
Update error returned for various HttpStatus codes

pallavijustice:
### Change description ###

Update error message returned for certain HttpStatus codes to ensure that Liberata contact information in not unnecessarily added to the message

https://tools.hmcts.net/jira/browse/NFDIV-4086


src/main/java/uk/gov/hmcts/divorce/payment/PaymentService.java
```diff
@@ -326,4 +330,12 @@ public Double getServiceCost(String service, String event, String keyword) {
 
         return feeResponse.getAmount();
     }
+
+    private boolean isGenericErrorRequiredForHttpStatus(HttpStatus httpStatus) {
+        boolean required = switch (httpStatus) {
+            case INTERNAL_SERVER_ERROR, GATEWAY_TIMEOUT -> true;
+            default -> false;
+        };
+        return required;
```
adamg-hmcts, line 339: How about using ||? There are only 2 conditions:
```suggestion
        return httpStatus == INTERNAL_SERVER_ERROR || httpStatus == GATEWAY_TIMEOUT;
```

pallavijustice: Had started with this but was wondering if in future we will need more custom error messages based on HttpStatus. If that could be the case, I could have this function return the error message itself based on value of httpStatus?
The ticket however mentions only 2 codes (500, 504)

adamg-hmcts (dismissed):

DelythJustice (dismissed):

src/main/java/uk/gov/hmcts/divorce/payment/PaymentService.java
```diff
@@ -186,6 +189,10 @@ private PbaResponse getPbaErrorResponse(String pbaNumber, FeignException excepti
             return new PbaResponse(httpStatus, String.format(NOT_FOUND.value(), pbaNumber), null);
         }
 
+        if (isGenericErrorRequiredForHttpStatus(httpStatus)) {
```
pallavijustice, line 192: Will triggered a scenario where the internal feign client for fee-and-pay isn't reachable. If that scenario is possible as well (which is likely), we will need to bring the check for the httpStatus(500, 504) codes right after the FeignException is caught to return the generic error before other conditions are checked for other possible error conditions.

src/main/java/uk/gov/hmcts/divorce/payment/PaymentService.java
```diff
@@ -186,6 +189,10 @@ private PbaResponse getPbaErrorResponse(String pbaNumber, FeignException excepti
...
+            return new PbaResponse(httpStatus, ERROR_GENERIC, pbaNumber);
+        }
+
```
pallavijustice, line 195: Will triggered a scenario where the internal feign client for fee-and-pay isn't reachable. If that scenario is possible as well (which is likely), we will need to bring the check for the httpStatus(500, 504) codes right after the FeignException is caught to return the generic error before other conditions are checked for other possible error conditions.

adamg-hmcts (approved):

DelythJustice (approved):

PR #3856, opened 2024-07-11
NFDIV 4055 court service validation error when international address selected

pallavijustice:
### Change description ###

When the service method is selected as 'court service' is below events when respondent is identified as having an overseas address, then system must not allow this change.
1. Solicitor submit application
2. Solicitor change service request
3. Caseworker change service request
4. Caseworker reissue application 

https://tools.hmcts.net/jira/browse/NFDIV-4055


src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorChangeServiceRequest.java
```diff
@@ -123,6 +126,13 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(
                 .data(caseData)
                 .errors(singletonList("You may not select Solicitor Service if the respondent is confidential."))
                 .build();
+        } else if (caseData.getApplicationType().isSole()
+            && application.isCourtServiceMethod()
+            && isOverseasApp2AndNotConfidentialContactDetails) {
+            return AboutToStartOrSubmitResponse.<CaseData, State>builder()
```
adamg-hmcts, line 132: Nit: Might be worth reusing your [validateCaseFieldsForCourtService](https://github.com/hmcts/nfdiv-case-api/pull/3856/files#diff-9b990846c4b9d80a24906609305e6f222dcb9d6a3250d56328ede2a459114015R204) method here, as that will keep all the business logic in one place will be easy to change if they ask us to adjust the rules in the future

adamg-hmcts (approved):

src/main/java/uk/gov/hmcts/divorce/solicitor/event/SolicitorChangeServiceRequest.java
pallavijustice: Have reused the same method now.

adamg-hmcts (approved):

PR #3884, opened 2024-07-19
NFDIV-4183. Add a label to create general email event

pallavijustice:
### Change description ###

Add a label before attachments to warn users that attachment size should be less than 2MB

https://tools.hmcts.net/jira/browse/NFDIV-4183


src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerGeneralEmail.java
```diff
@@ -96,6 +99,7 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
             .mandatory(GeneralEmail::getGeneralEmailOtherRecipientEmail, "generalEmailParties=\"other\"")
             .mandatory(GeneralEmail::getGeneralEmailOtherRecipientName, "generalEmailParties=\"other\"")
             .mandatory(GeneralEmail::getGeneralEmailDetails)
+            .label("attachmentWarning", WARNING_ATTACHMENT_SIZE)
             .optional(GeneralEmail::getGeneralEmailAttachments)
             .done();
     }
```
adamg-hmcts, line 105: We could also check whether there's an attachment and then show a warning on event submission:
image
If there's a warning in the callback response, it shows them the warning message and they have to confirm event submission by submitting again.

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerGeneralEmail.java
```diff
@@ -60,6 +60,9 @@ public class CaseworkerGeneralEmail implements CCDConfig<CaseData, State, UserRo
     private static final String NO_VALID_EMAIL_ERROR
         = "You cannot send an email because no email address has been provided for this party.";
 
+    private static final String WARNING_ATTACHMENT_SIZE = "### NOTE: Individual attachments must be less than 2MB"
```
adamg-hmcts, line 63: Nitpicky suggestion - would "WARNING" catch their attention more than  "NOTE"?
```suggestion
    private static final String WARNING_ATTACHMENT_SIZE = "### WARNING: Individual attachments must be less than 2MB"
```

adamg-hmcts (approved):

PR #3902, opened 2024-07-25
[NFDIV-4221] Set solicitor firm address during common component NoC

adamg-hmcts:
### Change description ###

Common component NoC update to start automatically populating the sol address and blanking out old sol details when the new sol is assigned to a case.

https://tools.hmcts.net/jira/browse/NFDIV-4221

**Before raising**

**Before merging**

### DIFF
--- a/src/main/java/uk/gov/hmcts/divorce/noticeofchange/service/ChangeOfRepresentativeService.java
+++ b/src/main/java/uk/gov/hmcts/divorce/noticeofchange/service/ChangeOfRepresentativeService.java
@@ -100,15 +100,16 @@ public class ChangeOfRepresentativeService {
         applicant.setOffline(YesOrNo.NO);
     }
 
-    private void updateOrgPolicyAndSolicitorDetails(Solicitor applicantSolicitor, OrganisationsResponse nocRequestUserOrg,
+    private void updateOrgPolicyAndSolicitorDetails(Solicitor applicantSolicitor, OrganisationsResponse nocRequestingUserOrg,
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/hmcts/divorce/noticeofchange/service/ChangeOfRepresentativeService.java
```diff
@@ -96,12 +100,18 @@ private void setApplicantRepresented(Applicant applicant) {
...
         applicantSolicitor.setName(String.join(" ", nocRequestingUser.getFirstName(), nocRequestingUser.getLastName()));
         applicantSolicitor.setEmail(loggedInUserEmail);
-        applicantSolicitor.setFirmName(nocRequestUserFirmName);
+        applicantSolicitor.setFirmName(nocRequestUserOrg.getName());
+        applicantSolicitor.setAddress(
+            parseOrganisationAddress(nocRequestUserOrg.getContactInformation())
+        );
+        applicantSolicitor.setReference(null);
+        applicantSolicitor.setAddressOverseas(null);
+        applicantSolicitor.setPhone(null);
```
adamg-hmcts, line 114: 3 fields will be null:
- **Reference:** Seems to be an optional field that is not usually populated.
- **AddressOverseas:** We could possibly write a method that works this out from the address, but as the address won't be used for an online sol and so far with international post we have required that they explicitly specify when their address is international, it seems safest to leave it as null until the sol updates their own contact details.
- **Phone:** Doesn't seem to be stored by the professional API - the sol can fill this in later if they update their contact details.

FaisalMoJ (approved):

PR #3916, opened 2024-07-30
NFDIV-4164 FO granted has no CO date on it 

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-4164


src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerExpediteFinalOrder.java
```diff
@@ -88,6 +88,13 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToStart(CaseDetails<Ca
         log.info("{} about to start callback invoked for Case Id: {}", CASEWORKER_EXPEDITE_FINAL_ORDER, details.getId());
         var caseData = details.getData();
 
+        if (caseData.getConditionalOrder().getGrantedDate() == null) {
+            return AboutToStartOrSubmitResponse.<CaseData, State>builder()
+                .data(caseData)
+                .errors(Collections.singletonList("No Conditional Order Granted Date found.  Unable to continue."))
```
adamg-hmcts, line 94: Nitpick: Might be worth putting the error message in a constant and then we can removed hardcoded strings from these events/tests and will only need to update it once if it changes in the future

adamg-hmcts (approved):

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerExpediteFinalOrder.java
BWeeks-JusticeGovUK: Good shout - I've done that, prefixing each constant with ERROR_ so they are easy to find with code completion.

adamg-hmcts (approved):

PR #3935, opened 2024-08-06
NFDIV-4237 Cron to resend AosResponseDocumentPack for affected cases

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-4237


src/main/java/uk/gov/hmcts/divorce/common/service/task/ResendJSCitizenAOSResponseLetters.java
```diff
@@ -0,0 +1,42 @@
...
+    private final ResendJudicialSeparationCitizenAosResponseNotification resendJudicialSeparationCitizenAosResponseNotification;
+
+    private final NotificationDispatcher notificationDispatcher;
+
+    @Override
+    public CaseDetails<CaseData, State> apply(CaseDetails<CaseData, State> details) {
+        final var data = details.getData();
+
+        try {
+            notificationDispatcher.send(resendJudicialSeparationCitizenAosResponseNotification, data, details.getId());
```
DelythJustice, line 32: I think you could have the flag check before the send rather than further in, but if you want extra logs confirming letter not resent then it's fine

DelythJustice (approved):

src/main/java/uk/gov/hmcts/divorce/systemupdate/schedule/SystemRegenerateJsCitizenAosResponseLetterTask.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    public void run() {
+        log.info("SystemRegenerateJsCitizenAosResponseCoverLetterTask started");
+        try {
+            final User user = idamService.retrieveSystemUpdateUserDetails();
+            final String serviceAuth = authTokenGenerator.generate();
+
+            List<Long> caseIds = taskHelper.loadCaseIds(CASE_ID_CSV);
+            final BoolQueryBuilder query = boolQuery()
+                .filter(QueryBuilders.termsQuery("reference", caseIds));
```
adamg-hmcts, line 62: How about filtering by the `JsCitizenAosResponseLettersResent` flag here or checking the flag before triggering the task on line 67? 🤔 It would allow us to avoid [this logging](https://github.com/hmcts/nfdiv-case-api/pull/3935/files#diff-a44b64cc6ce5cb2549f04f52d49a57cbb0c4c8d2a6fcfbb15ac1f71da5279aefR47)

BWeeks-JusticeGovUK: I thought it would be useful to be able to confirm if a something was getting skipped (just in case).

adamg-hmcts: I suppose a bigger problem with filtering at the end is we could put the event in case history multiple times when it does nothing. The extra logging and wasted work of calling the event/task multiple times probably aren't an issue if it's a small migration

BWeeks-JusticeGovUK: It's sub 60 cases this time, so very small.  I'll have a think about it anyway.  Can probably get the best of both by moving the log entry.

adamg-hmcts (dismissed):

DelythJustice (approved):

PR #3980, opened 2024-08-23
NFDIV-4243 CW requests for Information - Sole cases

BWeeks-JusticeGovUK:
Request for Information Feature Branch

Due to be merged on the 11/12/24


src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerRequestForInformationJoint.java
```diff
@@ -0,0 +1,119 @@
...
+    private final RequestForInformationNotification requestForInformationNotification;
+
+    private final NotificationDispatcher notificationDispatcher;
+
+    @Override
+    public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuilder) {
+        new PageBuilder(configBuilder
+            .event(CASEWORKER_REQUEST_FOR_INFORMATION_JOINT)
+            .forAllStates()
+            .name("Request For Information")
```
FaisalMoJ, line 53: This could be replaced with a constant as it is repeated multiple times

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerRequestForInformationJoint.java
```diff
@@ -0,0 +1,119 @@
...
+            .description("Request for information")
+            .showCondition("applicationType=\"jointApplication\"")
+            .showSummary()
+            .showEventNotes()
+            .endButtonLabel("Submit")
+            .aboutToSubmitCallback(this::aboutToSubmit)
+            .grant(CREATE_READ_UPDATE, CASE_WORKER)
+            .grantHistoryOnly(SUPER_USER, LEGAL_ADVISOR, JUDGE, SOLICITOR, CITIZEN, JUDGE))
```
FaisalMoJ, line 61: I believe citizen cannot view history via XUI so may be remove it from history list?

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/RequestForInformation.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @CCD(
+        label = "Email address",
+        typeOverride = Email,
+        access = {DefaultAccess.class}
+    )
+    private String requestForInformationEmailAddress;
+
+    @CCD(
+        label = "Secondary Name",
```
FaisalMoJ, line 53: Label casing is inconsistent i.e some starts with first word capital and rest lower case

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerRequestForInformationSole.java
```diff
@@ -0,0 +1,119 @@
...
+    private final RequestForInformationNotification requestForInformationNotification;
+
+    private final NotificationDispatcher notificationDispatcher;
+
+    @Override
+    public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuilder) {
+        new PageBuilder(configBuilder
+            .event(CASEWORKER_REQUEST_FOR_INFORMATION_SOLE)
+            .forAllStates()
+            .name("Request For Information")
```
FaisalMoJ, line 53: Same as the comment on joint class

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerRequestForInformationJoint.java
```diff
@@ -0,0 +1,119 @@
...
+            .page("requestForInformation", this::midEvent)
+            .pageLabel("Request For Information")
+            .complex(CaseData::getRequestForInformationList)
+                .complex(RequestForInformationList::getRequestForInformation)
+                    .mandatory(RequestForInformation::getRequestForInformationJointParties)
```
adamg-hmcts, line 66: We could set the options that will be available to choose from as the request for information parties in an aboutToStart or have 2 versions of the input with a show condition based on application type? It could help us avoid duplicating the event

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerRequestForInformationSole.java
```diff
@@ -0,0 +1,119 @@
...
+        if (!errors.isEmpty()) {
+            return AboutToStartOrSubmitResponse.<CaseData, State>builder()
+                .errors(errors)
+                .build();
+        }
+
+        return AboutToStartOrSubmitResponse.<CaseData, State>builder()
+            .data(details.getData())
+            .build();
+    }
```
adamg-hmcts, line 89: I think the notification client provides a useful error when information is invalid? Maybe we could reuse that/let it propagate from the aboutToSubmit callback rather than validating emails here and including validation logic in the helper

BWeeks-JusticeGovUK: Possibly?  This is essentially the same logic used for general emails.

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerRequestForInformationJoint.java
BWeeks-JusticeGovUK: If you setvup the list in aboutToStart you have to use a dynamic list, which doesn't work with show conditions, so you can't hide/show the text inputs on demand unfortunately.  A fixed list is the only way to handle those.

If you use multiple dropdowns, you need a valid show condition.  They need to be based on a field used in the event, and the event doesn't require applicationType, which means you end up with a visible field that isn't necessary.  I spent a while trying to work around that, but couldn't find an obvious solution.

adamg-hmcts: For a show condition to work, I was thinking we could get application type by using a readOnly field that's always hidden - we have similar config in CaseworkerOfflineDocumentVerified:
`.readonlyNoSummary(CaseData::getApplicationType, ALWAYS_HIDE)`

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/RequestForInformation.java
BWeeks-JusticeGovUK: Fixed

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerRequestForInformationJoint.java
BWeeks-JusticeGovUK: Fixed

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerRequestForInformationHelper.java
```diff
@@ -0,0 +1,167 @@
...
+        final RequestForInformation requestForInformation = caseData.getRequestForInformationList().getRequestForInformation();
+        final String emailAddress = isRepresented ? applicant.getSolicitor().getEmail() : applicant.getEmail();
+        final String name = isRepresented ? applicant.getSolicitor().getName() : applicant.getFullName();
+        if (TRUE.equals(setSecondary)) {
+            requestForInformation.setRequestForInformationSecondaryEmailAddress(emailAddress);
+            requestForInformation.setRequestForInformationSecondaryName(name);
+        } else {
+            requestForInformation.setRequestForInformationEmailAddress(emailAddress);
+            requestForInformation.setRequestForInformationName(name);
+        }
```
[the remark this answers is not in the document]
adamg-hmcts: Ah I haven't seen a design for the new tab and I was comparing to how we display information for general emails and NoC:
image
We show the target party that was entered in the event and then a caseworker could figure out the details of the target if they would like to. I was thinking about whether doing the same here could make the code simpler and reduce the chance of bugs, although I appreciate the extra detail could be useful

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerRequestForInformationJoint.java
BWeeks-JusticeGovUK: Resolved, per slack discussion.

adamg-hmcts (approved):

DelythJustice (approved):

adamg-hmcts (dismissed):

adamg-hmcts (dismissed):

adamg-hmcts (approved):

PR #3991, opened 2024-09-04
NFDIV-4303 Remove access to certain case tabs for respondent solicitor

pallavijustice:
### Change description ###

Remove access to History , conditional order and Outcome of conditional order tab for respondent's solicitor. 

https://tools.hmcts.net/jira/browse/NFDIV-4303


src/main/java/uk/gov/hmcts/divorce/divorcecase/tab/ApplicationHistoryTab.java
```diff
@@ -0,0 +1,37 @@
...
+
+import static uk.gov.hmcts.divorce.divorcecase.model.UserRole.APPLICANT_1_SOLICITOR;
+import static uk.gov.hmcts.divorce.divorcecase.model.UserRole.APPLICANT_2_SOLICITOR;
+import static uk.gov.hmcts.divorce.divorcecase.model.UserRole.CASE_WORKER;
+import static uk.gov.hmcts.divorce.divorcecase.model.UserRole.JUDGE;
+import static uk.gov.hmcts.divorce.divorcecase.model.UserRole.LEGAL_ADVISOR;
+import static uk.gov.hmcts.divorce.divorcecase.model.UserRole.SUPER_USER;
+
+@Component
+public class ApplicationHistoryTab implements CCDConfig<CaseData, State, UserRole> {
```
adamg-hmcts, line 18: Nitpick: Event history or case history might be more consistent with the naming we use in conversation with service?
```suggestion
public class EventHistoryTab implements CCDConfig<CaseData, State, UserRole> {
```

src/main/java/uk/gov/hmcts/divorce/divorcecase/tab/CaseTypeTab.java
```diff
@@ -475,6 +507,24 @@ private void buildOutcomeOfConditionalOrderTab(ConfigBuilder<CaseData, State, Us
...
+                    AwaitingClarification,
+                    AwaitingAmendedApplication,
+                    AwaitingPronouncement,
+                    ClarificationSubmitted,
+                    LAReview,
+                    SeparationOrderGranted
+                )
+            );
+        addOutcomeOfConditionalOrderTabFields(tabBuilder);
+    }
```
adamg-hmcts, line 526: Nitpick: If we move this above `addOutcomeOfConditionalOrderTabFields` it will be sat next to the equivalent method for applicant 1 solicitor and easier for someone to compare the two versions of the tab.

I also wonder whether we could pass the first part of the showCondition (`"applicationType=\"jointApplication\""`) into addOutcomeOfConditionalOrderTabFields() and put the duplicated show condition code in the shared method? e.g:

```
private void addOutcomeOfConditionalOrderTabFields(final Tab.TabBuilder<CaseData, UserRole> tabBuilder, String showCondition) {
  tabBuilder.showCondition(showCondition + " AND coGranted=\"*\" OR "
                    + showForState(
                    AwaitingAdminClarification,
                    AwaitingClarification,
                    AwaitingAmendedApplication,
                    AwaitingPronouncement,
                    ClarificationSubmitted,
                    LAReview,
                    SeparationOrderGranted
                )
            )
}
```

adamg-hmcts (dismissed):

src/main/java/uk/gov/hmcts/divorce/divorcecase/tab/ApplicationHistoryTab.java
pallavijustice: Had to name it so since the ccd config generator would otherwise pick up the ApplicationTab.java first and CaseTypeTab next and assign TabOrder of lower values to the tabs in these files. This means that History tab would be last.
The only other way around it to force a sequence in which ccd config generator would pickup files in this directory.

pallavijustice: Fiel renamed to ApplicationEventHistory.java

adamg-hmcts (dismissed):

adamg-hmcts (approved):

PR #4025, opened 2024-09-23
[NFDIV-4314] Create payment service requests in backend (Part 2)

adamg-hmcts:

https://tools.hmcts.net/jira/browse/NFDIV-4314

### Change description ###
Moves service request creation into the backend for citizen payments, to avoid multiple service requests being created when a payment does not work the first time.

### Old Frontend Payment Workflow:
Users begin citizen frontend payments by triggering PaymentPostController, which sends a series of requests to prepare the case for payment before the user inputs their card details.

**1.) If the case is not in AwaitingPayment or AwaitingFinalOrderPayment state:**
POST /CITIZEN_SUBMIT or POST /RESPONDENT_APPLY_FINAL_ORDER
- Changes case state to AwaitingPayment or AwaitingFinalOrderPayment.
- Saves payment order summary to the case data for the frontend to use.

**2.) POST F&P/service-request to create a service request.** 

**3.) POST F&P/service-request/card-payment to take payment using the service request.**

**Problem with this workflow:** Service requests are not saved to the case data immediately. If the user exits the frontend between steps 2 and 3, we don’t have any record of the service request and create another service request the next time they log in.

### New Frontend Payment Workflow:
The service request is now saved to case data at the start of the payment process, within CITIZEN_SUBMIT or RESPONDENT_APPLY_FINAL_ORDER.

**1.) If the case is not in AwaitingPayment or AwaitingFinalOrderPayment state:**
POST /CITIZEN_SUBMIT or POST /RESPONDENT_APPLY_FINAL_ORDER
- Changes case state to AwaitingPayment or AwaitingFinalOrderPayment.
- Saves payment order summary to the case data for the frontend to use.
- Saves service request reference to the case data for the frontend to use.

**2.) POST F&P/service-request/card-payment to take payment using the service request.**

**Backwards Compatibility:** Some cases will already be in AwaitingPayment or AwaitingFinalOrderPayment. To avoid resubmitting the citizen-submit and respondent-apply-final-order events, I’ve put in a temporary event (citizen-create-service-request) that doesn’t appear in case history and will be triggered to create the service request for these cases. We could remove this event in a tech debt shortly after the new workflow has been merged and all cases have progressed beyond AwaitingPayment/AwaitingFinalOrderPayment


src/functionalTest/java/uk/gov/hmcts/divorce/testutil/FunctionalTestSuite.java
```diff
@@ -120,7 +120,7 @@ private CaseDetails submitNewCase(
     protected Response triggerCallback(Map<String, Object> caseData, String eventId, String url) throws IOException {
 
         // Add random Long generation of 16 digits to avoid duplicate case id
-        long caseId = Math.abs(new Random().nextLong());
+        long caseId = ThreadLocalRandom.current().nextLong(1_000_000_000_000_000L, 10_000_000_000_000_000L);
```
adamg-hmcts, line 123: Changed this because the F&P service request endpoint validates that the caseId is 16 digits long and the previous use of Random could generate IDs of different lengths

DelythJustice: no harm in having that explainer as comment really

src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenSubmitApplication.java
```diff
@@ -105,5 +107,18 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(CaseDetails<C
             .build();
     }
 
+    public void setServiceRequestReferenceForApplicationPayment(CaseData data, long caseId) {
+        final Application application = data.getApplication();
+
```
DelythJustice, line 112: do this creates service request reference always?

adamg-hmcts: Yep unless F&P API fails to return one due to an error, in which case an exception will propagate from PaymentService and the user will have to try again later

FaisalMoJ (approved):

PR #4040, opened 2024-09-27
NFDIV-4329 Return error from Draft AoS if AoS already submitted

pallavijustice:
### Change description ###

Return error from Draft AoS if AoS already submitted

https://tools.hmcts.net/jira/browse/NFDIV-4329


src/main/java/uk/gov/hmcts/divorce/common/event/DraftAos.java
```diff
@@ -125,6 +125,10 @@ private List<String> validateDraftAos(final CaseData caseData) {
         final var acknowledgementOfService = caseData.getAcknowledgementOfService();
         final List<String> errors = new ArrayList<>();
 
+        if (null != acknowledgementOfService && null != acknowledgementOfService.getDateAosSubmitted()) {
+            errors.add("The Acknowledgement Of Service has already been submitted.");
```
adamg-hmcts, line 129: Nitpick: If we put this in a constant then we don't need to put the same hardcoded string in the test

adamg-hmcts (approved):

adamg-hmcts (approved):

PR #4084, opened 2024-10-16
NFDIV-4394 - Update notification to new sol when NOC is run

pallavijustice:
### Change description ###

Update notification to new sol when NOC is run to include additional text. Also include new webForm url instead of email address.

https://tools.hmcts.net/jira/browse/NFDIV-4394


DelythJustice (dismissed):
Just to point out Fubara hasn't yet added the AC for the web form link.

adamg-hmcts (approved):

PR #4121, opened 2024-10-29
NFDIV-4471 - Add welsh template for letter to citizen when NoC grants case access to Sol

pallavijustice:
### Change description ###

Add welsh template for letter to citizen when NoC grants case access to Sol

https://tools.hmcts.net/jira/browse/NFDIV-4471


pallavijustice:
Duplicate PR

PR #4141, opened 2024-11-06
NFDIV-3985 Send letter to applicant when new paper case is created

pallavijustice:
### Change description ###

When a new paper case is created via the bulk scanning process, a letter must be sent to the applicant confirming the receipt of application.

https://tools.hmcts.net/jira/browse/NFDIV-3985


src/main/java/uk/gov/hmcts/divorce/document/content/PaperApplicationReceivedTemplateContent.java
```diff
@@ -0,0 +1,58 @@
...
+public class PaperApplicationReceivedTemplateContent implements TemplateContent {
+
+    private final DocmosisCommonContent docmosisCommonContent;
+    private final Clock clock;
+
+    public static final String DATE_OF_RESPONSE = "dateOfResponse";
+    private static final int SUBMISSION_RESPONSE_DAYS = 28;
+
+    @Override
+    public List<String> getSupportedTemplates() {
```
DelythJustice, line 38: is this needed at all, I thought I saw it hard coded in another bit

DelythJustice: I think we have some code using DocumentGenerator and some not, I think you aren't using it so likely don't need this?

DelythJustice (approved):

adamg-hmcts (approved):

src/main/java/uk/gov/hmcts/divorce/caseworker/service/notification/PaperApplicationReceivedNotification.java
```diff
@@ -69,10 +78,35 @@ private Document generateDocument(final long caseId,
                                       final Applicant applicant,
                                       final CaseData caseData) {
 
+        boolean isApplicant1 = applicant == caseData.getApplicant1();
```
adamg-hmcts, line 80: It might be fine as it is, but I think we'd usually use .equals() here to do the check based on object equality (values of the instance vars due to Lombok @Data annotation) rather than references in memory?

adamg-hmcts (dismissed):

adamg-hmcts (approved):

PR #4145, opened 2024-11-08
NFDIV-4374: Add new template and opening times var

FaisalMoJ:
### Change description ###

Add new template and opening times var

https://tools.hmcts.net/jira/browse/NFDIV-4374

**Before raising**

**Before merging**


src/main/java/uk/gov/hmcts/divorce/notification/CommonContent.java
```diff
@@ -160,28 +163,32 @@ public Map<String, String> mainTemplateVars(final CaseData caseData,
         templateVars.put(SIGN_IN_URL, getSignInUrl(caseData));
         templateVars.put(WEBFORM_URL, config.getTemplateVars().get(WEBFORM_URL));
         templateVars.put(SMART_SURVEY, getSmartSurvey());
+
+        getPhoneAndOpeningTimes(applicant, templateVars);
         return templateVars;
     }
 
-    public Map<String, String> basicTemplateVars(final CaseData caseData, final Long caseId) {
+    public Map<String, String> basicTemplateVars(final CaseData caseData, final Long caseId, Applicant applicant) {
```
adamg-hmcts, line 171: Might be worth passing in language preference rather than applicant, as it could make the method more flexible in the future (e.g. we could pass in a sol language preference rather than an applicant language preference) and it would be easier to understand because we'd only pass in the args required and wouldn't have multiple variables named `applicant` in basicTemplateVars. Do you think it would take long to refactor?

adamg-hmcts (approved):

PR #4164, opened 2024-11-18
NFDIV-4512 new

DelythJustice:
### Change description ###

Enter a description.

https://tools.hmcts.net/jira/browse/NFDIV-4512

**Before raising**

**Before merging**


src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerFindMatches.java
```diff
@@ -83,39 +85,58 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToStart(final CaseDeta
...
-            .should(QueryBuilders.termQuery(field, name)) // Exact match
-            .should(QueryBuilders.termQuery(field, " " + name)) // Prepend space
-            .should(QueryBuilders.termQuery(field, name + " ")) // Append space
-            .should(QueryBuilders.termQuery(field, " " + name + " ")); // Prepend and append space
+    public String[] normalizeAndSplit(String name) {
+        // remove brackets and anything inside them
+        String nameWithoutStuffInBrackets = name.replaceAll("\\([^)]*\\)", "");
+
+        // replace multiple consecutive "/" with a single "/" because we might have multiple names due to translations
+        String nameWithoutExtraSlashes = nameWithoutStuffInBrackets.replaceAll("/+", "/");
```
adamg-hmcts, line 93: It's difficult to know what's best for this, as other characters might be used to separate translations 🤔. I was thinking it could make sense to remove all punctuation and then we can use a simple regex (`[^a-zA-Z0-9]`). Not sure whether that will cater to every scenario though

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerFindMatches.java
```diff
@@ -83,39 +85,58 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToStart(final CaseDeta
...
+
+        // remove illegal characters that we've spotted in prod data
+        String illegalCharacters = ".=!*_";
```
adamg-hmcts, line 96: There were some other punctuation characters in the DB - did you find it's safe to leave them in?
```
"-"
"'"
""""
"/"
"`"
"("
":"
"!"
","
"|"
" ​"
"_"
"]"
"="
")"
"."
"*"
```

DelythJustice: If they've put something in brackets, I'm assuming it isn't the name and the brackets and what's in the brackets gets removed.
The / seems to be to split alternative spellings - this is a bit of a weird one because they are meant to have exactly what is on the certificate which obvs isn't going to be 2 names, but I've gone with splitting on / and then doing extra searches on the array of names. Just based on the example you gave.
I thought I'd covered what you gave me but must not have if '`-_| are also in there. I did get my prod access renewed now tho. Maybe I need to do some more searching?
Trying to think if folk can have `' in name, I think they can have - in their name.

The alternative is to do fuzzy search but that would give more false positives I think.

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerFindMatches.java
DelythJustice: If I remove all punctuation I'm not going to be able to match on the other names they might have given.
Sion Corn / Santa Claus would then be Sion Corn Santa Claus so wouldn't match on either Sion Corn nor Santa Claus I think. 
One thing I did wonder about was splitting off according to any weird characters and then doing the matches on the shorter names - so might be doing match on first name, last name, foreign spelling first name, foreign spelling last name and it would give more false positives but all on matching marriage date only.
Then this example would match on Sion, Corn, Santa, Claus and also Sion Corn, Sion Siabod, Sian Sion etc.
What do you reckon?

DelythJustice: we are going with splitting stored name to array of names where there are illegal chars but also removing 'name changed by deed poll'

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerFindMatches.java
```diff
@@ -83,39 +86,46 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToStart(final CaseDeta
...
-    // Helper method to create term queries with variations for spaces
-    private BoolQueryBuilder createNameMatchQuery(String field, String name) {
-        return QueryBuilders.boolQuery()
-            .should(QueryBuilders.termQuery(field, name)) // Exact match
-            .should(QueryBuilders.termQuery(field, " " + name)) // Prepend space
-            .should(QueryBuilders.termQuery(field, name + " ")) // Append space
-            .should(QueryBuilders.termQuery(field, " " + name + " ")); // Prepend and append space
+    public String[] normalizeAndSplit(String name) {
+        // remove "name changed by Deed Poll" ignoring case
+        String nameWithoutDeedPollStatement = name.replaceAll("(?i)name changed by deed poll", "").trim();
```
adamg-hmcts, line 91: It's a bit of a nitpick, but I suppose we could have variations of this, e.g. "changed by deed poll". Another way we could do it is to look for any text in brackets that includes "deed poll" although it might not be in brackets and there's a risk we could strip out a real name if a bracket was included at the start of the input by mistake. Handling all the scenarios gets very complex and I'm not sure what's best

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerFindMatches.java
```diff
@@ -83,39 +86,46 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToStart(final CaseDeta
...
+
+        // split to separate names for search wherever there are non alphanumeric characters
+        return Arrays.stream(nameWithoutDeedPollStatement.split("[^a-zA-Z0-9]+"))
```
adamg-hmcts, line 94: I think this will remove non-ASCII characters which might be a valid part of the name, e.g. European names with accents ("é"), Polish names with ogoneks ("ę"), Chinese symbols and others. We could put in a weaker regex that only excludes punctuation or accept that some letters won't be matched if we think the impact will be low

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerFindMatches.java
DelythJustice: if they put in exactly as on uk marriage cert then this should be ok, if they don't then it will match against each word which likely won't hit a lot of names, maybe the 'by' and 'poll' with letby or pollard or somesuch.

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerFindMatches.java
DelythJustice: I've updated the code to cope not on the unicode and added to tests

adamg-hmcts (approved):

PR #4211, opened 2024-12-04
NFDIV-4343 Implementation of Case flags 2.1 for NFD

pallavijustice:
### Change description ###

Enable CaseFlags for NFD. Allow caseworkers to create case level and party level case flags.

https://tools.hmcts.net/jira/browse/NFDIV-4343


src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerCreateCaseFlag.java
```diff
@@ -0,0 +1,67 @@
...
+            .complex(CaseData::getPartyFlags)
+            .optional(PartyFlags::getApplicant1Flags, ALWAYS_HIDE, true, true)
+            .optional(PartyFlags::getApplicant2Flags, ALWAYS_HIDE, true, true)
+            .optional(PartyFlags::getApplicant1SolicitorFlags, ALWAYS_HIDE, true, true)
+            .optional(PartyFlags::getApplicant2SolicitorFlags, ALWAYS_HIDE, true, true)
+            .done()
+            .optional(CaseData::getFlagLauncher,
+                null, null, null, null, "#ARGUMENT(CREATE,VERSION2.1)");
+    }
+
```
adamg-hmcts, line 55: You might need to add a submitted callback to this event for the flag completion page

src/main/java/uk/gov/hmcts/divorce/noticeofchange/event/SystemApplyNoticeOfChange.java
```diff
@@ -103,6 +105,8 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToStart(final CaseDeta
 
         CaseData responseData = objectMapper.convertValue(data, CaseData.class);
 
+        caseFlagsService.resetSolicitorCaseFlags(responseData, isApplicant1);
```
adamg-hmcts, line 108: I would shift it above the API call that's done via `assignCaseAccessClient`, where we make the other data changes (e.g buildChangeOfRepresentative and resetConditionalOrderFields). 

If it's arranged in that order and there's some error in the code, the event will fail, case data will roll back to what it was originally, and the sol won't have been granted case access through AAC API

src/main/java/uk/gov/hmcts/divorce/systemupdate/service/CcdUpdateService.java
```diff
@@ -357,6 +360,18 @@ public void incrementOrgAssignedUsersSupplementaryData(final String caseId,
...
     }
 
+    public void submitSupplementaryDataToCcdForServiceID(final String caseId,
+                                                           final String authorisation,
+                                                           final String serviceAuth) {
+        log.info("submitSupplementaryDataToCcdForServiceID Case ID: {}", caseId);
+        Map<String, Map<String, Map<String, Object>>> supplementaryData = new HashMap<>();
+        supplementaryData.put("supplementary_data_updates",
+            singletonMap(SET_OPERATION,
+                singletonMap(SERVICE_ID_FIELD, SERVICE_ID)));
```
adamg-hmcts, line 370: This looks like it should work to avoid overwriting org information in the supplementary data. It might be worth us having a test case for Elis to confirm, as it could be a bad bug and time-consuming to fix if we did overwrite any org info by mistake

adamg-hmcts (approved):

PR #4221, opened 2024-12-09
[NFDIV-4564] Superuser event to remove case notes

adamg-hmcts:
### Change description ###

Enter a description.

https://tools.hmcts.net/jira/browse/NFDIV-4564

**Before raising**

**Before merging**


src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerRemoveNote.java
```diff
@@ -0,0 +1,65 @@
...
+            .optional(CaseData::getNotes)
+            .done();
+    }
+
+    public AboutToStartOrSubmitResponse<CaseData, State> midEvent(final CaseDetails<CaseData, State> details,
+                                                                       final CaseDetails<CaseData, State> beforeDetails) {
+
+        final var beforeNotes = beforeDetails.getData().getNotes();
+        final var notes = details.getData().getNotes();
+        boolean hasAddedNotes = notes != null && (beforeNotes == null || beforeNotes.size() < notes.size());
```
pallavijustice, line 52: It is hard to protect against but users could add and remove a note within this event and this condition wouldn't catch that.

pallavijustice (approved):

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerRemoveNote.java
adamg-hmcts: Thanks, I didn't think it was worth complicating the validation to avoid that, as we are only making it available to superusers and would also have to block them from editing notes, which they could find useful to redact pieces of information without fully removing a note

PR #4260, opened 2025-01-10
NFDIV-4319 - Set HmctsServiceId for new cases on case submission.

pallavijustice:
### Change description ###

Set HmctsServiceId for new cases on case submission. This is required for case flags feature.

https://tools.hmcts.net/jira/browse/NFDIV-4319


src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerHwfApplicationAccepted.java
```diff
@@ -66,6 +68,10 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(
         details.setData(caseworkerHwfApplicationAndPaymentHelper.setDateSubmittedAndDueDate(caseData));
         caseworkerHwfApplicationAndPaymentHelper.setRequiredCaseFieldsForPostSubmissionCase(details);
 
+        if (beforeDetails.getState() != State.Submitted && details.getState() == State.Submitted) {
```
adamg-hmcts, line 71: It can also go into "AwaitingDocuments", if the app didn't upload all their docs. I think you could set up supplementary data here regardless of the end state?

PR #4280, opened 2025-01-21
NFDIV-4632 - Initialise case flags structure in case data when case is submitted

pallavijustice:
### Change description ###

While testing NFDIV-4343 a null pointer exception was seen. This was because of some of the data being initialised in aboutToStart of createCaseFlags event did not persist when the event finished. With this change, initialisation of case flag data is being done at submission of a case. This is similar to NFDIV-4319 where HMCTSServiceID was being set when case was being submitted.

https://tools.hmcts.net/jira/browse/NFDIV-4632


src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerHwfApplicationAccepted.java
```diff
@@ -67,12 +67,13 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(
         log.info("{} about to submit callback invoked for Case Id: {}", CASEWORKER_HWF_APPLICATION_ACCEPTED, details.getId());
         CaseData caseData = details.getData();
 
+        caseFlagsService.initialiseCaseFlags(caseData);
         caseData.setCaseFlagsSetupComplete(YesOrNo.YES);
```
adamg-hmcts, line 71: I don't think you need to set the values here, as cases go through citizen-submit when they move to AwaitingHWFDecision. Probably no harm in leaving it as it is though

PR #4311, opened 2025-02-05
NFDIV-4601 Updating the states for Solicitor  "General Applicant" event is available

pallavijustice:
### Change description ###
NFDIV-4601 Updating the states for Solicitor  "General Applicant" event is available

https://tools.hmcts.net/jira/browse/NFDIV-4601


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/State.java
```diff
@@ -552,6 +552,17 @@ public enum State {
...
+    public static final EnumSet<State> GENERAL_APPLICATION_STATES = EnumSet.complementOf(EnumSet.of(
+        Draft,
+        AwaitingApplicant1Response,
+        AwaitingApplicant2Response,
+        Applicant2Approved,
+        AwaitingPayment,
+        Withdrawn,
+        Rejected,
+        Archived
+    ));
```
adamg-hmcts, line 564: Nitpick: You might be able to use `POST_SUBMISSION_STATES` like caseworker general application, as it's almost identical to this EnumSet? If it's going to be a custom set of states that's only for the sol general application event, I'd be tempted to list them in the event and leave out the new EnumSet

pallavijustice: Moved the EnumSet to the event rather than keeping it in State.java

adamg-hmcts (approved):

PR #4312, opened 2025-02-05
NFDIV-4660 Prevent AoS Tab showing fields if AoS is not submitted

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-4660


src/main/java/uk/gov/hmcts/divorce/divorcecase/tab/CaseTypeTab.java
```diff
@@ -95,6 +95,15 @@ public class CaseTypeTab implements CCDConfig<CaseData, State, UserRole> {
...
 
+    private static final String AOS_SUBMITTED = "dateAosSubmitted=\"*\"";
+    private static final String AOS_SUBMITTED_APP_2_ONLINE = "dateAosSubmitted=\"*\" AND applicant2Offline=\"No\"";
+    private static final String AOS_SUBMITTED_APP_2_OFFLINE = "dateAosSubmitted=\"*\" AND applicant2Offline=\"Yes\"";
+    private static final String AOS_SUBMITTED_NO_JURISDICTION = "dateAosSubmitted=\"*\" AND jurisdictionAgree=\"No\"";
+    private static final String AOS_SUBMITTED_APP_2_REPRESENTED = "dateAosSubmitted=\"*\" AND applicant2SolicitorRepresented=\"Yes\"";
+    private static final String AOS_SUBMITTED_APP_2_PRIVATE_REPRESENTED =
+        "dateAosSubmitted=\"*\" AND applicant2ContactDetailsType!=\"private\" AND applicant2SolicitorRepresented!=\"Yes\"";
+    private static final String AOS_SUBMITTED_STATEMENT_OF_TRUTH = "dateAosSubmitted=\"*\" AND statementOfTruth!=\"*\"";
+
```
adamg-hmcts, line 106: It's quite detailed - do you think it would be enough to use dateAosSubmitted as a tab show condition to address the incident and leave these other conditions out? It could keep the code simpler and extra conditions are potential bugs in the future if CTSC don't agree with them or we forget to update them after some future change to the events

BWeeks-JusticeGovUK: @adamg-hmcts The only change was adding dateAosSubmitted - the other conditions were already in place, so I don't think there should be any complaints from CTSC there.

adamg-hmcts: @BWeeks-JusticeGovUK Oh good, sorry I missed that. There's that tab showCondition on line 158 that we could update, as it looks like visibility of all the fields is dependent on dateAosSubmitted?

BWeeks-JusticeGovUK: @adamg-hmcts Yeah you're quite right!  Change made!

adamg-hmcts (approved):

PR #4368, opened 2025-03-05
NFDIV-4670 Add Validations in CW events to allow only valid characters in names

pallavijustice:
### Change description ###

Add Validations in CW events to allow only valid characters in names

https://tools.hmcts.net/jira/browse/NFDIV-4670


FaisalMoJ (dismissed):

src/test/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerAmendCaseTest.java
```diff
@@ -69,4 +69,45 @@ void shouldNotReturnErrorIfMarriageDateMoreThanOneYearPriorToApplicationSubmitte
...
+
+    @Test
+    void shouldReturnErrorIfMarriageDetailsApplicantNamesHaveInvalidCharacters() {
+        final CaseData caseData = caseDataWithMarriageDate();
+        caseData.getApplication().setDateSubmitted(LocalDateTime.now());
+        caseData.getApplication().getMarriageDetails().setDate(
+            LocalDate.from(caseData.getApplication().getDateSubmitted().minusYears(1).minusDays(1))
+        );
+        caseData.getApplication().getMarriageDetails().setApplicant1Name("Inva(id App1Name");
+        caseData.getApplication().getMarriageDetails().setApplicant2Name("Inva(id App2Name");
```
adamg-hmcts, line 81: ValidationUtil is responsible for the character checks and we could have a simple test in each of these classes to verify they delegate to ValidationUtil, then test format validations in detail in ValidationUtil tests

A downside of doing it like this, is when ValidationUtil changes, it can mean tests for many other classes also have to be adjusted

pallavijustice: How do you suggest I do that. Is there a way I can check that invocation of the static method within ValidationUtil?

adamg-hmcts (approved):

PR #4415, opened 2025-03-26
NFDIV-4777 - Updates for change in fees impacting NFD

pallavijustice:
### Change description ###

Enter a description.

https://tools.hmcts.net/jira/browse/NFDIV-4777


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/ConditionalOrderQuestions.java
```diff
@@ -103,7 +103,7 @@ public class ConditionalOrderQuestions {
     @CCD(
         label = "Do you need to change your application or add anything?",
         hint = "If you change or add anything which means your application has to be sent to your "
-            + "husband/wife again you may have to pay a £95 fee"
+            + "husband/wife again you may have to pay a £59 fee"
```
adamg-hmcts, line 106: Do you know where we use this hint? I'm not sure we show it anywhere and thinking it could be simpler to remove the fee amount, as it will mean we don't need to change this yearly and can release this code immediately

pallavijustice: It doesn't get used anywhere so I will ask Fubara if we can change the sentence.

pallavijustice: Fee amount removed from hint

adamg-hmcts (dismissed):

adamg-hmcts (approved):

PR #4419, opened 2025-03-28
NFDIV-4714 Applicant Options flow - Happy path

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-4714


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/CaseData.java
```diff
@@ -380,6 +380,13 @@ public class CaseData {
     @Builder.Default
     private PartyFlags partyFlags = new PartyFlags();
 
+    @JsonUnwrapped
+    @CCD(
+        label = "General Application Options",
+        access = {DefaultAccess.class}
+    )
+    private GeneralApplicationOptions generalApplicationOptions;
```
adamg-hmcts, line 388: When we get to both applicants being able to put in a general app, I think they could both have options to fill in and be doing it simultaneously. Thinking that to make sure we can reuse this, it could be safest to save it on Applicant.java - it's also owned by one specific citizen if that makes sense 🤔

We can set `searchable = false` with the newer version of config gen, to prevent fields being indexed and causing our ES index to get big. Might be worth doing here as Izzy suggested CTSC won't need this info and it seems unlikely we'll be using it in a search filter or cron

adamg-hmcts (approved):

PR #4453, opened 2025-04-24
NFDIV-4838 - Trigger general referral when solicitor triggers a switch to sole at FO stage

pallavijustice:
### Change description ###

Trigger general referral when solicitor triggers a switch to sole at FO stage and final order is overdue. This would make the functionality similar to when a citizen triggers a S2S and FO is overdue. 

https://tools.hmcts.net/jira/browse/NFDIV-4838


src/main/java/uk/gov/hmcts/divorce/solicitor/event/Applicant2SolicitorSwitchToSoleFo.java
```diff
@@ -52,11 +56,16 @@ public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuild
...
+            .aboutToSubmitCallback(this::aboutToSubmit)
+            .submittedCallback(this::submitted))
             .page("App2SolSwitchToSoleFO", this::midEvent)
             .pageLabel(APPLY_FOR_FINAL_ORDER)
+            .complex(CaseData::getFinalOrder)
+                .readonlyNoSummary(FinalOrder::getIsFinalOrderOverdue, "doesApplicant2WantToApplyForFinalOrder=\"NEVER_SHOW\"")
+            .done()
             .complex(CaseData::getFinalOrder)
                 .mandatory(FinalOrder::getDoesApplicant2WantToApplyForFinalOrder)
+                .mandatory(FinalOrder::getApplicant2FinalOrderExplanation, "isFinalOrderOverdue=\"Yes\"")
```
adamg-hmcts, line 68: Do you know why we need to set it? Thinking that it should be set automatically by the FO overdue cron. Also thinking there could be some scenario where they can switch to sole before the FO is overdue and the automatic gen referral shouldn't happen - by setting this we'd introduce a bug

adamg-hmcts: I think they do - just reviewed the citizen frontend

adamg-hmcts (dismissed):

PR #4463, opened 2025-05-01
NFDIV-4858 Solicitors should pay the Application Fee appropriate to the date of submission, not draft

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-4858


BWeeks-JusticeGovUK:
This approach isn't valid - we can't overwrite existing order summaries/refs - this will leave open entries with F&P and cause issues.

PR #4502, opened 2025-05-22
NFDIV-4733 Applicant Options flow - (UnHappy Path – In person service / Alternative service)

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-4733


src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenGenerateD10.java
```diff
@@ -0,0 +1,89 @@
...
+        log.info("{} aboutToSubmit callback invoked for Case Id: {}", CITIZEN_GENERATE_D10, details.getId());
+
+        details.getData().getApplication().setServiceMethod(PERSONAL_SERVICE);
+        CaseData data = caseTasks(generateD10Form).run(details).getData();
+
+        return AboutToStartOrSubmitResponse.<CaseData, State>builder()
+            .data(data)
+            .state(AwaitingService)
+            .build();
+    }
```
adamg-hmcts, line 88: Is it only the D10 we should generate or also the NoP and other docs? Initially I'd thought we'd be regenerating all of them to account for any changes in contact details and show conditional content that's based on service type 🤔

src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenGenerateD10.java
```diff
@@ -0,0 +1,89 @@
...
+    ));
+
+    @Override
+    public void configure(final ConfigBuilder<CaseData, State, UserRole> configBuilder) {
+
+        configBuilder
+            .event(CITIZEN_GENERATE_D10)
+            .forStates(CITIZEN_UPDATE_STATES)
+            .showCondition(NEVER_SHOW)
+            .name("Citizen Generate D10")
```
adamg-hmcts, line 70: Do you know if Fubara is happy the name will be clear to CTSC in event history? We seem to have a "Change service request" event for sols that's used to change the service type and regen the NoP. Something similar for citizens could possibly fit

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/NoResponseOwnSearches.java
```diff
@@ -0,0 +1,21 @@
...
+    YES("I have already tried to find my partner's contact details"),
+
+    @JsonProperty("no")
+    NO("I have not tried to find my partner's contact details"),
+
+    @JsonProperty("notFound")
+    NOT_FOUND("I've not been able to find contact details for my partner");
+
+    private final String label;
+}
```
adamg-hmcts, line 21: For Bailiff, I've had to put in a "Yes, No, or Not Known" enum. Thinking we could possibly reuse that here or have some other generic reusable enum for these answers, although fine to leave it as is if we feel there will be an issue with that

src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenGenerateD10.java
BWeeks-JusticeGovUK: Refactored - Will generate Application, Respondent NoP, and D10.

src/main/java/uk/gov/hmcts/divorce/citizen/event/CitizenGenerateD10.java
BWeeks-JusticeGovUK: Refactored - 'Generate Process Server Docs'

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/NoResponseOwnSearches.java
BWeeks-JusticeGovUK: As per our call, this is likely to come up a few times.  Might be best to leave and re-evaluate at a later point to see if we can condense multiple enums in this way.

PR #4551, opened 2025-06-18
NFDIV-4763 Dispense with service 1

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-4763


src/main/java/uk/gov/hmcts/divorce/divorcecase/model/InterimApplicationOptions.java
```diff
@@ -36,6 +36,13 @@ public class InterimApplicationOptions {
     )
     private DeemedServiceJourneyOptions deemedServiceJourneyOptions;
 
+    @JsonUnwrapped
+    @CCD(
+        label = "Dispense With Service Journey Options",
+        access = {DefaultAccess.class}
```
adamg-hmcts, line 42:
```suggestion
        label = "Dispense With Service Journey Options",
        access = {DefaultAccess.class},
        searchable = false
```
Worth us setting searchable to false unless we will need to use these fields in crons or search filters

adamg-hmcts (approved):

PR #4612, opened 2025-07-29
NFDIV-4849 Allow LA to make service decision on Alternative Service application

pallavijustice:
### Change description ###

Enter a description.

https://tools.hmcts.net/jira/browse/NFDIV-4849


src/main/java/uk/gov/hmcts/divorce/legaladvisor/event/LegalAdvisorMakeServiceDecision.java
```diff
@@ -167,14 +173,23 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(
...
                 }
             }
         }
 
         log.info("ServiceApplication decision. End State is {} Due date is {}", endState, caseDataCopy.getDueDate());
 
         log.info("Sending ServiceApplicationNotification case ID: {}", details.getId());
-        if (endState != ServiceAdminRefusal) {
+        if (endState != ServiceAdminRefusal
+            && !(ALTERNATIVE_SERVICE.equals(serviceApplication.getAlternativeServiceType()) && serviceApplication.isApplicationGranted())) {
```
adamg-hmcts, line 192: Did they want this condition to be left in, as I noticed we've updated the service application granted notification?

src/main/java/uk/gov/hmcts/divorce/legaladvisor/event/LegalAdvisorMakeServiceDecision.java
```diff
@@ -167,14 +173,23 @@ public AboutToStartOrSubmitResponse<CaseData, State> aboutToSubmit(
                         DEEMED_SERVICE_REFUSED,
                         SERVICE_REFUSAL_TEMPLATE_ID);
                     endState = AwaitingAos;
+                } else if (ALTERNATIVE_SERVICE.equals(serviceApplication.getAlternativeServiceType())) {
+                    generateAndSetOrderToDeemedOrDispenseDocument(
```
adamg-hmcts, line 177: Nit: It would be good to update this method name and the variable names within the method, to show it's not only for deemed and dispensed

src/main/java/uk/gov/hmcts/divorce/legaladvisor/event/LegalAdvisorMakeServiceDecision.java
pallavijustice: Yes. When an alternative service application is granted, thee ticket asked to push the case into "General Consideration Complete" and not generate the granted_document or send any notification.

I updated the template since it was harmless but could be used in case we want to send notification later.

PR #5179, opened 2026-07-10
NFDIV-4265 Identify unused docmosis templates

BWeeks-JusticeGovUK:
https://tools.hmcts.net/jira/browse/NFDIV-4265


src/integrationTest/resources/application-integration.yaml
```diff
@@ -0,0 +1,171 @@
+spring:
```
adamg-hmcts, line 1: Thinking it would be good to have a consistent approach for configuration of the functionals and contract tests, with an application-functional.yaml and application-contract.yaml

BWeeks-JusticeGovUK: Makes sense to me - I've updated the config for functional and contract tests.

adamg-hmcts (approved):

PR #5186, opened 2026-07-14
NFDIV-5633 - Delete General Referral Event

pallavijustice:
### Change description ###

Enter a description.

https://tools.hmcts.net/jira/browse/NFDIV-5633


src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerDeleteGeneralReferral.java
```diff
@@ -0,0 +1,122 @@
...
+import static uk.gov.hmcts.divorce.divorcecase.model.UserRole.SOLICITOR;
+import static uk.gov.hmcts.divorce.divorcecase.model.UserRole.SUPER_USER;
+import static uk.gov.hmcts.divorce.divorcecase.model.access.Permissions.CREATE_READ_UPDATE;
+
+@Slf4j
+@Component
+@RequiredArgsConstructor
+public class CaseworkerDeleteGeneralReferral implements CCDConfig<CaseData, State, UserRole> {
+    public static final String CASEWORKER_DELETE_GENERAL_REFERRAL = "delete-general-referral";
+    private static final String DELETE_GENERAL_REFERRAL = "Delete general referral";
```
adamg-hmcts: We use "Reject" as the verb for other deletions such as general referral and service application. An advantage would be that they appear next to each other in the event list.

Was the event name requested by Fubara?

src/main/java/uk/gov/hmcts/divorce/divorcecase/model/GeneralApplication.java
```diff
@@ -58,10 +58,10 @@ public class GeneralApplication {
     private LocalDateTime generalApplicationReceivedDate;
 
     @CCD(
-        label = "General application referral date"
+        label = "General application referred on"
```
adamg-hmcts: We should be able to set searchable to false for this field, unless it will be used in a cron or search filter?

pallavijustice: Will do

src/main/java/uk/gov/hmcts/divorce/caseworker/event/CaseworkerDeleteGeneralReferral.java
pallavijustice: Fubara hasn't mentioned the event specifically apart from the line below
"When a caseworker selects the delete referral event".

I think we could name it with "Reject" as you suggested.

adamg-hmcts (approved):

